PR #3, opened 2025-10-08
add pre-commit hook and dev requirements

subpath:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
```
subpath, line 2: Config is the same as in the any-llm repo: https://github.com/mozilla-ai/any-llm/blob/main/.pre-commit-config.yaml

I didn't add mypy at the moment, I think it will make a big git diff so we could add it in a separate PR imho

PR #112, opened 2026-03-10
feat: remove use-play-integrity header for access tokens

noahpodgurski:
### What's new

- Remove the required use-play-integrity-header when using access tokens


noahpodgurski:
This was made in case we want to merge it, but I explained to the Android team it's more optimal to pass the header so MLPA doesn't have to compute both on every request. I'll close for now

PR #129, opened 2026-03-24
feat: add play integrity & access token metrics

noahpodgurski:
- Add prometheus metric tracking for play integrity and access token auth access
[GENAI-4327](https://mozilla-hub.atlassian.net/browse/GENAI-4327)


subpath (approved):
Suggestion: It would be good to add tests for these new metrics 🙏

PR #130, opened 2026-03-26
Draft: feat: add chat_id to metrics

noahpodgurski:
## What's new

- Add passed chat-id header to prometheus metrics

TODO: Research glean vs sink https://mozilla-hub.atlassian.net/browse/GENAI-3967

https://mozilla.slack.com/archives/C09KKBSD0SG/p1769005992138379


src/mlpa/core/prometheus_metrics.py
```diff
@@ -145,7 +145,7 @@ class PrometheusMetrics:
     chat_completion_latency=Histogram(
         "mlpa_chat_completion_latency_seconds",
         "Chat completion latency in seconds.",
-        ["result", "model", "service_type", "purpose"],
+        ["result", "model", "service_type", "purpose", "chat_id"],
```
subpath: This is a critical thing, we should not push high cardinality filters to Prometheus

This should be tracked through logging -> log sink -> BigQuery or with server side Glean

PR #134, opened 2026-04-09
feat: add model cost region override map

noahpodgurski:
## What's new

- Added `scripts/patch-cost-map/patch-model-cost-map.sh` for automatically pulling the latest model map and patching based off `scripts/patch-cost-map/region_overrides.json`
- Ran script and uploaded for `export LITELLM_MODEL_COST_MAP_URL=...` on LiteLLM deployment


scripts/patch-cost-map/patch-model-cost-map.sh
```diff
@@ -0,0 +1,42 @@
...
+require_cmd jq
+
+if [[ ! -f "${OVERRIDES_FILE}" ]]; then
+  echo "Missing overrides file: ${OVERRIDES_FILE}"
+  exit 1
+fi
+
+echo "Preparing LiteLLM model cost map override..."
+curl -fsSL \
+  https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json \
```
subpath: Suggestion: 
I think we should be pulling from the release version

noahpodgurski: Good call

noahpodgurski:
Closed in favor of inline deployment yaml overrides

PR #138, opened 2026-04-15
feat: implement stream disconnect method AIPLAT-617

subpath:
Jira: [AIPLAT-617](https://mozilla-hub.atlassian.net/browse/AIPLAT-617)

## What's new:

1. Prometheus.ABORTED value to record partial tokens - this will need Grafana updates
2. catches aborted streaming requests

## QA

1. Tests old and new ✅ 
2. Local QA ✅ 

image

aborted metrics are recorded

image

### DIFF
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -1,3 +1,5 @@
+import asyncio
+import contextlib
 import json
 import time
 
@@ -229,6 +231,20 @@ async def stream_completion(
     logger.debug(
         f"Starting a stream completion using {authorized_chat_request.model}, for user {authorized_chat_request.user}",
     )
+
+    disconnect_event = asyncio.Event()
+    _client_disconnected_msg = (
+        f"Client disconnected mid-stream for user {authorized_chat_request.user}"
+    )
+
+    async def _watch_disconnect() -> None:
+        while True:
+            message = await request._receive()
+            if message.get("type") == "http.disconnect":
+                disconnect_event.set()
+                return
+
+    watch_task = asyncio.create_task(_watch_disconnect())
     try:
         client = get_http_client()
         async with client.stream(
@@ -281,11 +297,9 @@ async def stream_completion(
             litellm_routing_snapshot = parse_litellm_routing_headers(response.headers)
 
             async for chunk in response.aiter_bytes():
-                if await request.is_disconnected():
+                if disconnect_event.is_set():
                     result = PrometheusResult.ABORTED
-                    logger.info(
-                        f"Client disconnected mid-stream for user {authorized_chat_request.user}"
-                    )
+                    logger.info(_client_disconnected_msg)
                     break
 
                 if is_first_token:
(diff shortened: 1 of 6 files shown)


src/mlpa/core/completions.py
```diff
@@ -227,6 +231,20 @@ async def stream_completion(authorized_chat_request: AuthorizedChatRequest):
...
+    )
+
+    async def _watch_disconnect() -> None:
+        while True:
+            message = await request._receive()
+            if message.get("type") == "http.disconnect":
+                disconnect_event.set()
+                return
+
+    watch_task = asyncio.create_task(_watch_disconnect())
```
noahpodgurski, line 247: thought: I'd be interested to know how much latency or CPU processing this adds to the stream, hopefully it's negligable

src/mlpa/core/completions.py
```diff
@@ -376,6 +400,12 @@ async def stream_completion(authorized_chat_request: AuthorizedChatRequest):
         else:
             logger.error(f"Upstream service returned an error: {e}")
     finally:
+        watch_task.cancel()
+        with contextlib.suppress(asyncio.CancelledError):
+            await watch_task
```
noahpodgurski, line 407: question: what do these lines do?

src/mlpa/core/prometheus_metrics.py
```diff
@@ -7,6 +7,7 @@
 class PrometheusResult(StrEnum):
     SUCCESS = "success"
     ERROR = "error"
+    ABORTED = "aborted"
```
noahpodgurski: nit: I think `ABORT = "abort"` would be more consistent English, since we already have `ERROR` not `ERRORED`, but I don't mind either way

src/mlpa/run.py
```diff
@@ -176,7 +177,7 @@ async def chat_completion(
 
     if authorized_chat_request.stream:
         return StreamingResponse(
-            stream_completion(authorized_chat_request),
+            stream_completion(authorized_chat_request, request),
```
noahpodgurski, line 180: thought: `authorized_chat_request` is already sort of a sub-type of `request` I wonder if there's a way to pass in what we need through that object instead

noahpodgurski (approved):
Looks good just left some thoughts and nitpicks :)

src/mlpa/core/completions.py
subpath: yes, there is no pulling loop, it's I/O bound once per request

src/mlpa/core/completions.py
subpath: it's awaiting for a canceled task to be cleaned-up

src/mlpa/core/prometheus_metrics.py
subpath: Thanks, I agree 👍

src/mlpa/run.py
subpath: these are not the same thing authorized_chat_request is our Pydantic model and request is a starlette object

PR #139, opened 2026-04-16
fix:silent-streaming-timeouts AIPLAT-638

subpath:
## What's new: 
 Three cases where stream_completion would return an empty response with no error signal to the client:

 1.  Exception mid-stream - if LiteLLM dropped the connection after sending the first chunk, the generator stopped silently.
 2. Empty 200 body - if LiteLLM returned 200 with no body at all, the generator returned cleanly with zero chunks yielded.
 3.  Pool timeout override -  passing timeout=int to client.stream() silently overrode(is this the right past tense? 🤔 ) all httpx timeout phases including pool (5s -> 300s), causing a saturated connection pool to hang for 5 minutes instead of failing fast.

  All three now yield an error SSE frame so the client gets an actionable signal instead of an empty response. Also removed a dead outer except httpx.HTTPStatusError handler - all HTTP status errors are caught by the inner handler before the stream loop and could never reach it.

See slack thread: https://mozilla.slack.com/archives/C09AR1FJCJ1/p1776264568704779

## QA: 
So I did it TDD style : ) I've made 3 tests - one for each possible root-cause and they were all failing, then after fixes all tests are passing

### DIFF
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -252,7 +252,12 @@ async def stream_completion(
             LITELLM_COMPLETIONS_URL,
             headers=LITELLM_COMPLETION_AUTH_HEADERS,
             json=body,
-            timeout=env.STREAMING_TIMEOUT_SECONDS,
+            timeout=httpx.Timeout(
+                read=env.STREAMING_TIMEOUT_SECONDS,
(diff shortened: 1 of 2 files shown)


noahpodgurski (approved):
Looks good to me 👍

PR #140, opened 2026-04-17
fix: user lookup on the first call

subpath:
LiteLLM logs noisy ERRORs on every new user's first request:

 ```
"litellm.proxy.management_endpoints.customer_endpoints.end_user_info(): Exception occured - "
```

It was because end_user_info() catches its own 404 (user not found) and re-logs it, but message is always empty.

Now we check DB directly before calling /customer/info. 

QA: 
unit tests ✅ 
Local QA ✅

### DIFF
--- a/src/mlpa/core/utils.py
+++ b/src/mlpa/core/utils.py
@@ -40,14 +40,9 @@ async def get_or_create_user(user_id: str):
     claimed_new_identity = False
     try:
         params = {"end_user_id": user_id}
-        response = await client.get(
-            f"{env.LITELLM_API_BASE}/customer/info",
-            params=params,
-            headers=LITELLM_MASTER_AUTH_HEADERS,
-        )
-        user = response.json()
+        db_user = await litellm_pg.get_user(user_id)
 
-        if not user.get("user_id"):
+        if db_user is None:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/mlpa/core/utils.py [hunk 1]
```diff
@@ -40,14 +40,9 @@ async def get_or_create_user(user_id: str):
     claimed_new_identity = False
     try:
         params = {"end_user_id": user_id}
-        response = await client.get(
-            f"{env.LITELLM_API_BASE}/customer/info",
-            params=params,
-            headers=LITELLM_MASTER_AUTH_HEADERS,
-        )
-        user = response.json()
+        db_user = await litellm_pg.get_user(user_id)
```
noahpodgurski, line 42: thought: Curious on the implications this will have with connections made to the DB, I guess since we're pooling connections impact should be minimal

noahpodgurski (approved):

src/mlpa/core/utils.py
subpath: the pool is already used on every auth flow for capacity checks, so this adds one more fetchrow per request on the same pool and no new connections.

PR #141, opened 2026-04-22
feat: /v1/search with exa-search

noahpodgurski:
## What's new

- Added `/v1/search` support across the proxy by introducing `SearchRequest` and `AuthorizedSearchRequest` in , plus shared auth flow extraction in `src/mlpa/core/auth/authorize.py` so chat and search requests reuse the same authorization logic via `authorize_chat_request` and `authorize_search_request`.
- Extended configuration for search in src/mlpa/core/config.py and
    litellm_config.yaml, including EXA_API_KEY, LITELLM_SEARCH_URL, renamed
    virtual auth headers to LITELLM_VIRTUAL_AUTH_HEADERS, and a new search entry
    in user_feature_budget / valid_service_types.
- Added a new /v1/search/ route in src/mlpa/run.py and a corresponding
    get_search proxy path in src/mlpa/core/completions.py to forward authorized
    search requests to LiteLLM.
- Cap `SearchRequest.max_results` in `src/mlpa/core/classes.py`(1-10)


src/mlpa/core/completions.py
```diff
@@ -546,3 +551,34 @@ async def get_completion(authorized_chat_request: AuthorizedChatRequest):
...
+        f"Starting a search request using for user {authorized_search_request.user}",
+    )
+    try:
+        client = get_http_client()
+        response = await client.post(
+            f"{LITELLM_SEARCH_URL}/exa-search",
+            headers=LITELLM_VIRTUAL_AUTH_HEADERS,
+            json=body,
+        )
+        data = response.json()
```
ti3x, line 588: This looks like a real behavior change: we never call `response.raise_for_status()` here before parsing the JSON. That means a LiteLLM `429`, `400`, or `5xx` will come back from MLPA as a normal `200` response with an error-shaped body, so clients lose the actual failure status and rate-limit handling. I think this should mirror the completion path and raise on non-2xx before returning the payload.

src/mlpa/run.py
```diff
@@ -184,6 +185,24 @@ async def chat_completion(
...
+    request: Request,
+    authorized_search_request: Annotated[
+        AuthorizedSearchRequest, Depends(authorize_search_request)
+    ],
+):
+    if authorized_search_request.service_type != "search":
+        raise HTTPException(
+            status_code=400, detail="service-type header must be of type 'search'"
+        )
+    return await get_search(authorized_search_request)
```
ti3x, line 215: TODO: I get that this may just be for quick testing, but this route currently bypasses the usual user provisioning + blocked-user check from the chat path. If this endpoint is going to stick around, I think it should get the same gating before proxying.

noahpodgurski: Just added that, thanks

src/mlpa/core/completions.py
noahpodgurski: Fixed this as well

ti3x (approved):

PR #145, opened 2026-04-29
fix: exclude none

noahpodgurski:
## What's new

- Fix `exclude_unset` in `authorize.py` should be `exclude_none`


noahpodgurski:
https://github.com/mozilla/dataservices-infra/pull/1882

PR #150, opened 2026-05-11
fix: stream disconnect

noahpodgurski:
## What's new

- Fix stream disconnect issue by aborting upstream reads cleanly on client disconnect and avoiding spurious `ReadError` empty-response failures, especially before the first token.

### DIFF
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -352,7 +352,8 @@ async def stream_completion(
                 except StopAsyncIteration:
                     break
                 except httpx.ReadError:
-                    if disconnect_event.is_set():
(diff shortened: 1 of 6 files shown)


src/mlpa/core/completions.py
```diff
@@ -50,6 +51,8 @@
     ),
 }
 
+_DISCONNECT_POLL_INTERVAL_SECONDS = 0.1
```
subpath: Suggestion: This could be part of the config

subpath (approved):
Great ❤️ 
I think in the tests I've made for stream abortion I should have use  httpx.ReadError instead of the RuntimeError

PR #151, opened 2026-05-11
feat: search latency metric

noahpodgurski:
## What's new

- Add search latency metric to `/v1/search` requests
- Bump version to `1.0.1`
- Add mlpa version to `/readiness` response

### DIFF
--- a/src/mlpa/core/prometheus_metrics.py
+++ b/src/mlpa/core/prometheus_metrics.py
@@ -68,10 +68,13 @@ BUCKETS_LITELLM_ATTEMPTS = (0, 1, 2, 3, 5, float("inf"))
 
 @dataclass
 class PrometheusMetrics:
+    # global request metrics
     in_progress_requests: Gauge
     requests_total: Counter
     response_status_codes: Counter
     request_latency: Histogram
+
+    # auth
     validate_challenge_latency: Histogram
     validate_app_attest_latency: Histogram
     validate_app_assert_latency: Histogram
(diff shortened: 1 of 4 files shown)


src/mlpa/core/prometheus_metrics.py
```diff
@@ -216,6 +226,12 @@ class PrometheusMetrics:
         "Number of chat requests rejected due to budget, rate limit, payload size, or managed-user signup cap.",
         ["reason", "model", "service_type", "purpose"],
     ),
+    search_latency=Histogram(
+        "mlpa_search_latency_seconds",
+        "Search latency in seconds.",
+        ["result", "service_type", "purpose"],
+        buckets=BUCKETS_COMPLETION,
```
subpath: Suggestion: maybe make separate bucket list, I think exa search is much faster than LLM completion

src/mlpa/core/prometheus_metrics.py
```diff
@@ -216,6 +226,12 @@ class PrometheusMetrics:
         "Number of chat requests rejected due to budget, rate limit, payload size, or managed-user signup cap.",
         ["reason", "model", "service_type", "purpose"],
     ),
+    search_latency=Histogram(
+        "mlpa_search_latency_seconds",
+        "Search latency in seconds.",
+        ["result", "service_type", "purpose"],
```
subpath: does service_type and purpose makes sense here? 🤔

subpath (approved):

src/mlpa/core/prometheus_metrics.py
noahpodgurski: Maybe not, I figured we'd use them down the road, but I suppose we can add them back in at that point

PR #155, opened 2026-05-28
chore: refactoring-of-errors-and-metrics-handling

subpath:
completions.py file was huge, before adding new error's wrappers I've split the file into into smaller modules:
  - `metrics.py`: all metric recorders (chat and search)
  - `errors.py`: upstream error classification
  - `search.py`: search proxy (was mixed with chat code)

## QA: 
unit tests ✅


src/mlpa/run.py
```diff
@@ -16,7 +16,6 @@
 from mlpa.core.completions import (
     get_completion,
     get_or_create_user_for_completion,
```
noahpodgurski, line 18: Can we remove this import?

src/mlpa/run.py
```diff
@@ -206,9 +207,7 @@ async def search(
             status_code=400,
             detail={"error": "User not found from authorization response."},
         )
-    user, _ = await get_or_create_user_for_completion(
-        user_id, authorized_search_request
-    )
+    user, _ = await get_or_create_user(user_id)
```
noahpodgurski: I think we want to keep `get_or_create_user_for_completion` here - otherwise we'll miss out on the user cap logic

src/mlpa/core/completions.py
```diff
@@ -7,98 +7,48 @@
 import httpx
```
noahpodgurski, line 7: Can we bump the version in `pyproject.yaml` please?

src/mlpa/core/errors.py
```diff
@@ -0,0 +1,86 @@
...
+            return ERROR_CODE_RATE_LIMIT_EXCEEDED
+    except (json.JSONDecodeError, AttributeError, UnicodeDecodeError):
+        pass
+    if is_litellm_upstream_rate_limit(error_text):
+        return ERROR_CODE_UPSTREAM_RATE_LIMIT_EXCEEDED
+    return None
+
+
+def classify_upstream_error(
+    *,
```
noahpodgurski, line 63: What is `*,`?

src/mlpa/core/completions.py
```diff
@@ -414,50 +241,17 @@ async def _read_next_chunk(
...
+                tool_calls_accum[i] for i in sorted(tool_calls_accum)
+            )
+            record_completion_success(
+                authorized_chat_request,
+                prompt_tokens=prompt_tokens,
+                completion_tokens=completion_tokens,
+                tool_names=tool_names,
+                snapshot=litellm_routing_snapshot,
+            )
+            result = PrometheusResult.SUCCESS
```
noahpodgurski, line 257: I think having `record_completion_success` goes against the flow of storing the result in the `result` variable and passing that into the function. Can we do `record_completion_result` and pass in `result`? Or do we not want to record the latency of failed results?

src/mlpa/core/errors.py
subpath: it forces to use named argument

src/mlpa/core/completions.py
subpath: `record_completion_success` only fires on the happy path. Latency is already recorded with result in the finally block with `record_completion_latency`, if I understood your question correctly 🙏

noahpodgurski (approved):
Nice work 👍

PR #156, opened 2026-05-28
refactoring metrics patching

subpath:
## Switch unit tests to a real Prometheus registry

Added build_metrics(registry) factory + a metrics_spy pytest fixture that runs production code against a fresh CollectorRegistry per test. Migrated all metric-touching tests off MagicMock.

 Because MagicMock silently accepts label typos and locks tests to call-shape, they can break on any refactor. Real metrics catch typos at call time. New spy.assert_only({...}) fails on stray writes; tests assert on
  emitted values, not internal call shape.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mlpa"
-version = "1.0.6"
+version = "1.0.7"
 description = "Authenticates and proxies LLM requests through LiteLLM to enact budgets and per-user management."
 authors = [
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mlpa"
-version = "1.0.6"
+version = "1.0.8"
```
noahpodgurski: I think we can keep on 1.0.6 since we haven't released that yet

noahpodgurski (approved):

PR #159, opened 2026-05-28
Feat:  add new rejection wrappers

subpath:
Classify two more 400 cases coming back from LiteLLM so we don't lump them with generic errors:
  - error 7: invalid model name (client requested a model not configured in LiteLLM)
  - error 8: malformed request body (Vertex INVALID_ARGUMENT, generic LiteLLM BadRequestError)
  
  
  there are more than 100+ such errors per day, they are affecting our Success rate calculation


src/mlpa/core/errors.py
```diff
@@ -83,4 +87,19 @@ def classify_upstream_error(
             http_status=413,
             log_message=f"Context window exceeded for user {user}",
         )
+    if status_code == 400:
+        if is_invalid_model_name_error(error_text):
+            return RejectionMatch(
+                reason=PrometheusRejectionReason.INVALID_MODEL_NAME,
+                error_code=ERROR_CODE_INVALID_MODEL_NAME,
+                http_status=400,
+                log_message=f"Invalid model name for user {user}: {error_text}",
```
noahpodgurski, line 96: nit: The `413` `RejectionMatch.log_message` doesn't include the `error_text` - could we add it there or remove it here, just have it be consistent 🙏

noahpodgurski (approved):

PR #161, opened 2026-06-02
feat: override content-type for POST /verify/play

noahpodgurski:
## What's new

- Force `application/json` content type for `POST /verify/play` 
- This fixes a bug where our Android clients were sending the wrong content-type, resulting in 422 after the `fastapi==0.133.0` uptick
- Middleware adds on average ~1ms latency


subpath (approved):
Nice 👍 
Maybe add a test for it 👍

PR #163, opened 2026-06-02
chore(update-swagger-documentation)

subpath:
update missing error codes in the Swagger


src/mlpa/core/routers/play/play.py
```diff
@@ -103,7 +103,12 @@ def _validate_integrity_payload(payload: dict, expected_hash: str) -> None:
         raise HTTPException(status_code=401, detail="Device integrity check failed")
 
 
-@router.post("/play", tags=["Play Integrity"])
+@router.post(
+    "/play",
+    tags=["Play Integrity"],
```
noahpodgurski, line 108: Can we also check the 422 response in the docs for this? It should be 401 when the auth fails we should make that explicit I'm not sure where the 422 is being defined

noahpodgurski (approved):

PR #167, opened 2026-06-03
update error codes bad model, bad request

noahpodgurski:


noahpodgurski:
Updating error codes context: https://mozilla.slack.com/archives/C08PM2ULQ07/p1780929172033089

src/mlpa/core/config.py
subpath: It's not correct here too 🙏

subpath (approved):

PR #168, opened 2026-06-08
add `make docs` to precommit

noahpodgurski:
## What's new

- Add `make docs` to `lint and test`
- Remove model default value from request


.github/workflows/linters_and_tests.yml
```diff
@@ -36,3 +36,9 @@ jobs:
 
       - name: Run integration tests
         run: uv run pytest src/tests/integration -v
+
+      - name: Build docs
+        run: make docs
```
subpath, line 41: Question: how will it be commited?

subpath (approved):

.github/workflows/linters_and_tests.yml
noahpodgurski: It has to be commited in the MR, but `lint-and-test` verify docs step will fail if they don't match

PR #173, opened 2026-06-12
feat: Instrument pre-completion chat availability

randy-concepcion:
## What's New

Adds pre-completion coverage to `mlpa_chat_availability_total`, so auth rejections and user-provisioning failures show up in the chat availability signal.

A chat request terminates at one of four points and records exactly one outcome:

1. Auth (`authorize_chat_request`) - all excluded:
    - `invalid_service_type_for_model`
    - `auth_rejected`
    - `invalid_auth_request` 
2. User provisioning (`get_or_create_user_for_completion`):
    - `signup_cap_exceeded` (excluded)
    - `provisioning_failure` (failure, 5xx only)
3. Blocked check: `blocked` (excluded)
4. Model call (`get_completion` / `stream_completion`): already instrumented

## Notes

- The auth wrapper catches only `HTTPException` and records only 401 and 400. Everything else, including App Attest's explicit 500, is re-raised without recording.
- Shared-call 400s use `invalid_auth_request` rather than `invalid_purpose` because malformed App Attest base64 reaches the same path as purpose validation.
- `auth_system_failure` is defined but not emitted. See the limitation below.
- **Reference**: [Availability RFC](https://docs.google.com/document/d/1jo5BK1Im4i890qp-Wkey70hXpT80mV9NblaKoXLsGs0/edit?tab=t.0#heading=h.yh1fy3an43l) and [SLI Inventory](https://mozilla-hub.atlassian.net/wiki/spaces/FAAMT/pages/2745893471/Cloud+Services+Reliability+Inventory).

## Known Limitation

This is an interim measurement. For FxA and Play Integrity, the same `401` response is returned whether the request was correctly rejected (such as an expired token) or the authentication system itself had a problem (such as an outage). Since they can't be distinguished, all of these are counted as expected rejections and left out of the ratio for now.

This means that real authentication-system failures are not yet counted as failures. This is a known gap, already noted in the [Availability RFC](https://docs.google.com/document/d/1jo5BK1Im4i890qp-Wkey70hXpT80mV9NblaKoXLsGs0/edit?tab=t.0#heading=h.yh1fy3an43l) and the [SLI Inventory](https://mozilla-hub.atlassian.net/wiki/spaces/FAAMT/pages/2745893471/Cloud+Services+Reliability+Inventory), and it will be addressed once those failures can be identified separately.

The availability ratio is `success / (success + failure)`, and `excluded` and `abort` are held out.

### DIFF
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -27,7 +27,7 @@ def _resolve_purpose(service_type_value: str, purpose_header: str | None) -> str
     """Validate purpose header and return value; raise HTTPException if invalid."""
     valid = env.valid_purposes_for_service_type(service_type_value)
     requires = env.service_type_requires_purpose(service_type_value)
-    purpose = (purpose_header or "").strip()
+    purpose = purpose_header or ""
 
     if purpose:
         if purpose not in valid:
@@ -129,7 +129,7 @@ async def authorize_chat_request(
             AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
             model=chat_request.model,
             service_type=service_type.value,
-            purpose=(purpose or "").strip(),
+            purpose=purpose or "",
         )
         raise HTTPException(
             status_code=400,
@@ -161,7 +161,6 @@ async def authorize_chat_request(
         # - anything else (e.g. App Attest's explicit 500): re-raised unrecorded;
         #   auth-system-failure capture is left to a follow-on auth backend change
         # Non-HTTPException errors are not caught here and propagate unrecorded.
-        # Purpose is unknown at this point, so "" is always used as a placeholder.
         if exc.status_code == 401:
             reason = AvailabilityReason.AUTH_REJECTED
         elif exc.status_code == 400:
@@ -172,7 +171,7 @@ async def authorize_chat_request(
             reason,
             model=chat_request.model,
             service_type=service_type.value,
-            purpose="",
+            purpose=purpose or "",
         )
         raise
 
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -154,15 +154,14 @@ async def authorize_chat_request(
             use_play_integrity=use_play_integrity,
         )
     except HTTPException as exc:
-        # Record only the terminal auth dispositions the dependency intentionally
-        # emits. 401 is an expected or normalized auth rejection. 400 is a
-        # client-error request to the auth layer (invalid purpose, or malformed App
-        # Attest base64 decoded in app_attest_auth). Any other status, including App
-        # Attest's explicit 500, is re-raised unrecorded; auth-system-failure capture
-        # is left to a follow-on auth backend change. Non-HTTPException errors are not
-        # caught here and propagate unrecorded. Purpose is unresolved at these
-        # dispositions, so a stable "" placeholder is used, never the unvalidated
-        # header value.
+        # Only record terminal HTTP failures from the shared auth call:
+        # - 401: expected or normalized auth rejection (bad creds, expired token, etc.)
+        # - 400: client error to the auth layer (invalid purpose, or malformed App
+        #        Attest base64 decoded in app_attest_auth before its try block)
+        # - anything else (e.g. App Attest's explicit 500): re-raised unrecorded;
+        #   auth-system-failure capture is left to a follow-on auth backend change
+        # Non-HTTPException errors are not caught here and propagate unrecorded.
+        # Purpose is unknown at this point, so "" is always used as a placeholder.
         if exc.status_code == 401:
             reason = AvailabilityReason.AUTH_REJECTED
         elif exc.status_code == 400:
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -129,6 +129,8 @@ async def authorize_chat_request(
             AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
             model=chat_request.model,
             service_type=service_type.value,
+            # `purpose` is not yet resolved here
+            # This serves as a placeholder until a validated purpose is available
             purpose="",
         )
         raise HTTPException(
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -129,9 +129,7 @@ async def authorize_chat_request(
             AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
             model=chat_request.model,
             service_type=service_type.value,
-            # `purpose` is not yet resolved here
-            # This serves as a placeholder until a validated purpose is available
-            purpose="",
+            purpose=(purpose or "").strip(),
         )
         raise HTTPException(
             status_code=400,
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -27,7 +27,7 @@ def _resolve_purpose(service_type_value: str, purpose_header: str | None) -> str
     """Validate purpose header and return value; raise HTTPException if invalid."""
     valid = env.valid_purposes_for_service_type(service_type_value)
     requires = env.service_type_requires_purpose(service_type_value)
-    purpose = purpose_header or ""
+    purpose = (purpose_header or "").strip()
 
     if purpose:
         if purpose not in valid:
--- a/src/mlpa/core/auth/authorize.py
+++ b/src/mlpa/core/auth/authorize.py
@@ -13,6 +13,8 @@ from mlpa.core.classes import (
     ServiceType,
 )
 from mlpa.core.config import env
+from mlpa.core.metrics import record_chat_availability_for
+from mlpa.core.prometheus_metrics import AvailabilityReason
 from mlpa.core.routers.appattest import app_attest_auth
 from mlpa.core.utils import extract_user_from_play_integrity_jwt, parse_app_attest_jwt
 
@@ -123,27 +125,57 @@ async def authorize_chat_request(
     )
 
     if not is_service_type_valid:
+        record_chat_availability_for(
+            AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
+            model=chat_request.model,
+            service_type=service_type.value,
+            purpose="",
+        )
         raise HTTPException(
             status_code=400,
             detail=f"Invalid service-type value for model {chat_request.model}. Should be one of {env.forced_model_service_type_pairs.get(chat_request.model)}",
         )
 
-    return await _authorize_common_request(
-        request=request,
-        build_authorized_request=lambda user, purpose_value: AuthorizedChatRequest(
-            user=user,
+    try:
+        return await _authorize_common_request(
+            request=request,
+            build_authorized_request=lambda user, purpose_value: AuthorizedChatRequest(
+                user=user,
+                service_type=service_type.value,
+                purpose=purpose_value,
+                **chat_request.model_dump(exclude_unset=True),
+            ),
+            authorization=authorization,
+            service_type=service_type,
+            purpose=purpose,
+            x_dev_authorization=x_dev_authorization,
+            use_app_attest=use_app_attest,
+            use_qa_certificates=use_qa_certificates,
+            use_play_integrity=use_play_integrity,
+        )
+    except HTTPException as exc:
+        # Record only the terminal auth dispositions the dependency intentionally
+        # emits. 401 is an expected or normalized auth rejection. 400 is a
+        # client-error request to the auth layer (invalid purpose, or malformed App
+        # Attest base64 decoded in app_attest_auth). Any other status, including App
+        # Attest's explicit 500, is re-raised unrecorded; auth-system-failure capture
+        # is left to a follow-on auth backend change. Non-HTTPException errors are not
+        # caught here and propagate unrecorded. Purpose is unresolved at these
+        # dispositions, so a stable "" placeholder is used, never the unvalidated
+        # header value.
+        if exc.status_code == 401:
+            reason = AvailabilityReason.AUTH_REJECTED
+        elif exc.status_code == 400:
+            reason = AvailabilityReason.INVALID_AUTH_REQUEST
+        else:
+            raise
+        record_chat_availability_for(
+            reason,
+            model=chat_request.model,
             service_type=service_type.value,
-            purpose=purpose_value,
-            **chat_request.model_dump(exclude_unset=True),
-        ),
-        authorization=authorization,
-        service_type=service_type,
-        purpose=purpose,
-        x_dev_authorization=x_dev_authorization,
-        use_app_attest=use_app_attest,
-        use_qa_certificates=use_qa_certificates,
-        use_play_integrity=use_play_integrity,
-    )
+            purpose="",
+        )
+        raise
 
 
 async def authorize_search_request(
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -20,6 +20,7 @@ from mlpa.core.litellm_routing import parse_litellm_routing_headers
 from mlpa.core.logger import logger
 from mlpa.core.metrics import (
     extract_tool_names,
+    record_chat_availability,
     record_chat_request_rejection,
     record_completion_latency,
     record_completion_success,
@@ -27,6 +28,7 @@ from mlpa.core.metrics import (
     record_ttft,
 )
 from mlpa.core.prometheus_metrics import (
+    AvailabilityReason,
     PrometheusRejectionReason,
     PrometheusResult,
 )
@@ -80,6 +82,7 @@ async def stream_completion(
     record_request_with_tools(authorized_chat_request)
     body = _build_litellm_body(authorized_chat_request, stream=True)
     result = PrometheusResult.ERROR
+    availability_reason = AvailabilityReason.UPSTREAM_ERROR
     is_first_token = True
     prompt_tokens = 0
     completion_tokens = 0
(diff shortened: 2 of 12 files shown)


randy-concepcion:
Manual verification done in `dev`,  in addition to the unit tests.

**Verified manually**
- The three auth-stage excluded reasons, triggered with crafted requests and confirmed recording with the right labels and `outcome="excluded"`
  - `auth_rejected` (valid service-type, invalid auth)
  - `invalid_service_type_for_model` (valid model with a service-type it does not allow)
  - `invalid_auth_request` (invalid purpose)
- Real dev traffic exercised `success` and `upstream_error`, so the availability ratio is live, currently around 99% with dips that line up with `upstream_error` clusters
- Confirmed the counter is scraped into Prometheus and queryable

**Not verified manually**
`provisioning_failure`, `signup_cap_exceeded`, and `blocked` need specific user or system state, so they were not triggered in dev. They are covered by the unit tests. `auth_system_failure` is defined but not emitted, which is the known auth limitation in the description.

**Dashboard**
Test dashboard built in Yardstick to validate the metric and compare against the existing success-rate panels.
[[TEST] MLPA Availability](https://yardstick.mozilla.org/d/rclhslx/test-mlpa-availability?orgId=1&from=2026-06-12T22:40:02.064Z&to=2026-06-13T04:40:02.064Z&timezone=browser&var-environment=nonprod&var-datasource=edq6thuke248we&refresh=1m)

image

randy-concepcion:
I've also added the same panels to our [main Grafana dashboard](https://yardstick.mozilla.org/d/aolvm2k/llm-proxy-prod?orgId=1&from=now-3h&to=now&timezone=browser&var-environment=prod&var-datasource=adpvtjmrxoc1sb&refresh=1m) so once these changes are deployed, we can start seeing them in Prod.
* [MLPA Chat Availability (Rate)](https://yardstick.mozilla.org/d/aolvm2k/llm-proxy-prod?orgId=1&from=now-3h&to=now&timezone=browser&var-environment=prod&var-datasource=adpvtjmrxoc1sb&refresh=1m&viewPanel=panel-165)
* [MLPA Chat Availability (Time Series)](https://yardstick.mozilla.org/d/aolvm2k/llm-proxy-prod?orgId=1&from=now-3h&to=now&timezone=browser&var-environment=prod&var-datasource=adpvtjmrxoc1sb&refresh=1m&viewPanel=panel-166)
* [Exclusions/Failures Table](https://yardstick.mozilla.org/d/aolvm2k/llm-proxy-prod?orgId=1&from=now-3h&to=now&timezone=browser&var-environment=prod&var-datasource=adpvtjmrxoc1sb&refresh=1m&viewPanel=panel-167)

**MLPA Availability Panels (Prod)** - No data available until deployed.
image

src/mlpa/core/prometheus_metrics.py
```diff
@@ -19,6 +19,69 @@ class PrometheusRejectionReason(StrEnum):
...
+class AvailabilityReason(StrEnum):
+    # Strings shared with PrometheusRejectionReason are kept identical so the
+    # two counters reconcile. Keep them in sync when a rejection reason is added.
+
+    # --- completion-stage reasons (recorded inside stream_completion / get_completion) ---
+    CLEAN_COMPLETION = "clean_completion"  # success
+    UPSTREAM_ERROR = "upstream_error"  # failure
+    EMPTY_RESPONSE = "empty_response"  # failure
+    BUDGET_EXCEEDED = "budget_exceeded"  # excluded
+    RATE_LIMITED_OWN = "rate_limited_own"  # excluded
```
noahpodgurski: What is rate limited own?

src/mlpa/core/prometheus_metrics.py
```diff
@@ -19,6 +19,69 @@ class PrometheusRejectionReason(StrEnum):
...
+    RATE_LIMITED_UPSTREAM = "rate_limited_upstream"  # excluded
+    PAYLOAD_TOO_LARGE = "payload_too_large"  # excluded
+    INVALID_MODEL_NAME = "invalid_model_name"  # excluded
+    INVALID_REQUEST = "invalid_request"  # excluded
+    CLIENT_DISCONNECT = "client_disconnect"  # abort
+
+    # --- pre-completion reasons (recorded in the auth dependency and route body) ---
```
noahpodgurski, line 33: nit: Could we put this block above the completion-stage reasons just for organization's sake

src/mlpa/core/completions.py
```diff
@@ -52,20 +54,31 @@ def _build_litellm_body(req: AuthorizedChatRequest, *, stream: bool) -> dict:
 async def get_or_create_user_for_completion(
     user_id: str, req: AuthorizedChatRequest | AuthorizedSearchRequest
 ):
-    """Wraps get_or_create_user and records a signup-cap rejection metric if applicable."""
+    """
+    Wraps get_or_create_user and records availability for chat requests:
+    - signup cap (403 + MAX_USERS_REACHED): excluded, alongside the existing rejection metric
+    - user-resolution failure (status >= 500): failure
```
noahpodgurski: nit: status >= 500 should be server or system failure

src/mlpa/core/completions.py
```diff
@@ -52,20 +54,31 @@ def _build_litellm_body(req: AuthorizedChatRequest, *, stream: bool) -> dict:
...
+    - search requests and non-signup-cap, non-5xx failures: not recorded
+    """
     try:
         return await get_or_create_user(user_id)
     except HTTPException as exc:
-        if (
-            exc.status_code == 403
-            and isinstance(exc.detail, dict)
-            and exc.detail.get("error") == ERROR_CODE_MAX_USERS_REACHED
-            and isinstance(req, AuthorizedChatRequest)
```
noahpodgurski, line 63: question: Was this check unnecessary? Why did we remove it?

src/mlpa/core/completions.py
```diff
@@ -256,6 +272,7 @@ async def _read_next_chunk(
                 snapshot=litellm_routing_snapshot,
             )
             result = PrometheusResult.SUCCESS
+            availability_reason = AvailabilityReason.CLEAN_COMPLETION
```
noahpodgurski: nit: Thoughts on also just calling this `SUCCESS`?

src/mlpa/core/auth/authorize.py
```diff
@@ -123,27 +125,56 @@ async def authorize_chat_request(
     )
 
     if not is_service_type_valid:
+        record_chat_availability_for(
+            AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
+            model=chat_request.model,
+            service_type=service_type.value,
+            purpose="",
```
noahpodgurski: question: Why not pass the `chat_request.purpose` here?

src/mlpa/core/auth/authorize.py
```diff
@@ -123,27 +125,56 @@ async def authorize_chat_request(
...
-        use_app_attest=use_app_attest,
-        use_qa_certificates=use_qa_certificates,
-        use_play_integrity=use_play_integrity,
-    )
+            purpose="",
+        )
+        raise
 
 
 async def authorize_search_request(
```
noahpodgurski, line 179: thought: Can we add `record_chat_availability` to `authorize_search_request` too? At that point it might help to put into a common function too

src/mlpa/core/errors.py
```diff
@@ -18,6 +18,15 @@
     is_rate_limit_error,
 )
 
+_REJECTION_TO_AVAILABILITY_REASON: dict[
+    PrometheusRejectionReason, AvailabilityReason
+] = {
+    PrometheusRejectionReason.BUDGET_EXCEEDED: AvailabilityReason.BUDGET_EXCEEDED,
+    PrometheusRejectionReason.PAYLOAD_TOO_LARGE: AvailabilityReason.PAYLOAD_TOO_LARGE,
```
noahpodgurski, line 25: suggestion: We should also add tracking this to our middleware functions. For instance if `request_size.check_request_size_middleware` returns 413 `record_chat_availability` with `PAYLOAD_TOO_LARGE` should also fire

src/mlpa/core/prometheus_metrics.py
noahpodgurski: Maybe let's call it `rate_limited_user` instead?

noahpodgurski (changes requested):
Looks really good 🙌 just a few comments :)

src/mlpa/core/auth/authorize.py
randy-concepcion: Good question. I found the `chat_request` object didn't carry the `purpose` field since it's a `ChatRequest`, but saw it in the `AuthorizedChatRequest` object (sub-class of `ChatRequest`).

I figured we should leave `purpose` set to `""` as a placeholder until the request has been validated. I can add a comment here to make this more clear.

src/mlpa/core/auth/authorize.py
randy-concepcion: Ah, good thought here, I had to think about this for a bit :) I like the idea of leveraging how we defined chat completion availability for search. However, there is a risk of conflating what we measure for them. I think they should be distinct definitions, esp if search would have different handling and instrumentation.

I also like your thought on keeping things DRY with a common function, though I'd opt to defer this until we have at least one more SLI defined (i.e., for search). This way, that can better inform us on what info we can abstract out to a common function.

I'd def like to hear your thoughts on this esp if I'm missing something here!

src/mlpa/core/completions.py
randy-concepcion: Gotcha! I'll clarify this in the comment (as well as another comment that was elsewhere). Thanks for flagging this!

src/mlpa/core/completions.py
randy-concepcion: Good question. I think the diff made it look like it was removed, but I kept this check and moved it outside to cover both the existing `403` and new `500` status code checks.

The original signup cap rejection still triggers under the same conditions as before, and now also records the availability metric alongside it. The new `500` branch  records the availability metric for provisioning failure.

src/mlpa/core/completions.py
randy-concepcion: Thinking about this, I agree, `CLEAN_COMPLETION` isn't exactly clear. This might be a counter-nit, but how about `VALID_RESPONSE`? :-D In my mind (at least for me), this distinguishes from the `AvailabilityOutcome.SUCCESS`, as well as in the RFC doc where it mentions having "valid responses".

src/mlpa/core/errors.py
randy-concepcion: Ooo, I did not think of this. Great catch! I'll capture the `PAYLOAD_TOO_LARGE` for that check. One thing I noticed was that the middleware runs before the body is parsed and before auth, which means the `model`, `service_type`, and `purpose` aren't available yet. I'll set the `""` placeholders for those and pass the `reason` and `outcome` through.

src/mlpa/core/prometheus_metrics.py
randy-concepcion: Yes, good call on renaming this. I meant for this to differentiate rate limiting errors between our "own" platform (`rate_limited_own`) and 3rd party providers (`rate_limited_upstream`).

I think `user` feels like it could be the user causing the rate limiting, but would you be opposed to renaming to something like `rate_limited_platform` or `rate_limited_proxy` for naming clarity? I'm open to other suggestions!

src/mlpa/core/prometheus_metrics.py
randy-concepcion: Yes, good call. I'm all for better organization :) I also moved the corresponding reasons in the `_AVAILABILITY_OUTCOME_BY_REASON` map.

randy-concepcion (commented):
Thanks for the review @noahpodgurski! I replied to each comment inline. I'll make some of these changes now and adjust according to your responses.

src/mlpa/core/prometheus_metrics.py
randy-concepcion: Similar to the other comment, I just used `rate_limited_platform` for now to do some testing, but can adjust for something else!

src/mlpa/core/completions.py
randy-concepcion: I pushed up `VALID_RESPONSE` for now so I can run some tests, but can amend the commit if we want a different name!

src/mlpa/core/prometheus_metrics.py
noahpodgurski: Works for me! 👍

src/mlpa/core/completions.py
noahpodgurski: Sounds good 👍

src/mlpa/core/auth/authorize.py
noahpodgurski: That function has a `purpose` header parameter we could use - Or are you saying since it hasn't been officially resolved with `_resolve_purpose` we shouldn't pass it, since the value could be anything.

I guess I'm fine with either :)

src/mlpa/core/errors.py
noahpodgurski: Actually all of those values are available under `request.headers` and `request.body`. I think it'd be a good idea to pass those in - or were you thinking to explicitly exclude them since the request hasn't yet been verified?

src/mlpa/core/auth/authorize.py
randy-concepcion: Ahh, yeah, you're right about the `purpose` parameter that was passed in, but your second point was exactly what I was thinking: "the value could be anything". So if we used the unresolved purpose, we'd be labeling the metric with a value that hasn't been validated yet, so it could be something arbitrary.

noahpodgurski: Hmm actually now that I think about it more, I think it makes more sense to always pass it. This data could be helpful in making correlations later on. Also, in this particular case, we're already passing the invalid service type, so if we're already passing invalid values then I think purpose should also be included. We can always add whitelisting filtering down the road. WDYT?

randy-concepcion: Gotcha. Yeah, those are fair points, the correlation value is worth having. Let me update it to pass purpose in. 👍

src/mlpa/core/errors.py
randy-concepcion: You're right! I think one thing we may want to avoid is reading in the whole body to get the `model` though since we're rejecting with the content-length check. Having `service_type` and `purpose` could still be useful info to have, so we can still pass those through.

randy-concepcion:
@noahpodgurski Thanks again for the feedback! I've pushed changes and is ready for another round of review when you have a chance! :pray:

randy-concepcion:
Manually verified on `dev`. [Grafana link](https://yardstick.mozilla.org/d/rclhslx/test-mlpa-availability?orgId=1&from=2026-06-17T04:00:00.000Z&to=2026-06-18T03:59:59.000Z&timezone=browser&var-environment=nonprod&var-datasource=edq6thuke248we&refresh=1m) for 2026-06-17 time window.

**Prometheus: Metrics Explorer**
image


**MLPA Availability in Grafana** (includes failures)
image

src/mlpa/core/auth/authorize.py
```diff
@@ -123,27 +125,56 @@ async def authorize_chat_request(
     )
 
     if not is_service_type_valid:
+        record_chat_availability_for(
+            AvailabilityReason.INVALID_SERVICE_TYPE_FOR_MODEL,
+            model=chat_request.model,
+            service_type=service_type.value,
+            purpose=(purpose or "").strip(),
```
noahpodgurski: nit: Why the `.strip()`?

src/mlpa/core/auth/authorize.py
```diff
@@ -123,27 +125,56 @@ async def authorize_chat_request(
...
-        ),
-        authorization=authorization,
-        service_type=service_type,
-        purpose=purpose,
-        x_dev_authorization=x_dev_authorization,
-        use_app_attest=use_app_attest,
-        use_qa_certificates=use_qa_certificates,
-        use_play_integrity=use_play_integrity,
-    )
+            purpose="",
```
noahpodgurski: Can we pass `purpose` here too?

src/mlpa/core/middleware/request_size.py
```diff
@@ -19,6 +21,16 @@ async def check_request_size_middleware(request: Request, call_next):
...
                     )
+                    # `model` is in the request body, which we don't read here.
+                    # We reject on the Content-Length header without parsing it.
+                    record_chat_availability_for(
+                        AvailabilityReason.PAYLOAD_TOO_LARGE,
+                        model="",
+                        service_type=(
+                            request.headers.get("service-type") or ""
+                        ).strip(),
+                        purpose=(request.headers.get("purpose") or "").strip(),
```
noahpodgurski: nit: why the `strip()`?

src/mlpa/core/auth/authorize.py
randy-concepcion: Ah shoot. I put these in first initially and meant to remove them :facepalm: I've removed it.

src/mlpa/core/auth/authorize.py
randy-concepcion: Good catch! I added the purpose here too.

src/mlpa/core/middleware/request_size.py
randy-concepcion: Similar to above. I've removed it. :pray:

src/mlpa/core/auth/authorize.py
```diff
@@ -25,7 +27,7 @@ def _resolve_purpose(service_type_value: str, purpose_header: str | None) -> str
     """Validate purpose header and return value; raise HTTPException if invalid."""
     valid = env.valid_purposes_for_service_type(service_type_value)
     requires = env.service_type_requires_purpose(service_type_value)
-    purpose = (purpose_header or "").strip()
+    purpose = purpose_header or ""
```
noahpodgurski: Let's leave this line the way it was actually 👍

randy-concepcion: Makes sense to me, good call! 🛡️ Reverted back.

noahpodgurski (approved):

PR #174, opened 2026-06-12
fix: (the-unpaired-surrogates-crash): AIPLAT-942

subpath:
## Summary

Fixes [AIPLAT-942](https://mozilla-hub.atlassian.net/browse/AIPLAT-942). The `memories` service type was sitting at **~85% success rate** on the completion dashboard. The entire gap was one bug: requests containing a **lone/unpaired UTF-16 surrogate** crashed before reaching LiteLLM and returned a 502.

## Root cause

The OpenAI JS client occasionally truncates conversation text mid-emoji (~7KB in), leaving a dangling surrogate like `\ud83e` (half of 🧠/🤖). httpx serializes the outbound body with `ensure_ascii=False` and then `.encode("utf-8")`, which refuses to encode a lone surrogate:

```
'utf-8' codec can't encode character '\ud83e' ... surrogates not allowed
```

The exception was caught by the generic handler in `completions.py` → `raise_and_log(..., 502, "Failed to proxy request")`, and recorded as `result="error"`. Confirmed in prod: over 72h, **every** memories error (~22/day) was this same `UnicodeEncodeError`, accounting for the full 15% gap. The request never left MLPA — it's malformed input, not a model/backend issue.

## Changes

- New `src/mlpa/core/sanitization.py` module that collects sanitization methods

- Used for input and output sanitization in completion and search endpoints


.github/workflows/auto-bump-version.yml
```diff
@@ -114,6 +114,10 @@ jobs:
         if: steps.version.outputs.bump_needed == 'true'
         run: uv pip install --no-cache-dir -e .
 
+      - name: Update lockfile with new version
```
subpath, line 117: @noahpodgurski  I've also added a pre-commit hook to regenerate uv.lock to keep it up to date

noahpodgurski: saw that, thank you 🙏

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mlpa"
-version = "1.0.10"
```
noahpodgurski, line 7: FYI my version bump workflow is merged so we don't have to manually bump the version anymore, we should be able to merge the automated chore PR, and then make a release each time

subpath: ah, nice!

noahpodgurski (approved):

PR #177, opened 2026-06-16
feat: add Postgres query timeouts (AIPLAT-921)

subpath:
## Context

This PR is a prevention measure for a recent DB incident.

On 2026-04-30 our database ran out of connections for about an hour. Active
queries went from a normal ~3.5 up to 217, total connections hit 251, and the DB
started refusing new ones. The same thing happened in smaller bursts on 05-24,
05-25 and 05-26.

One of the reason is that our connection pool has no query timeout. So one slow query just
runs forever, holds its connection, and everything else piles up behind it until
the pool is full. This PR stops that.

## What's new

Two timeouts on the asyncpg pool, set on the Postgres side so the query is killed
even when the client is stuck:

- `statement_timeout = 3s`: a runaway query gets killed after 3s. This is the one
  that actually fixes Apr 30, it stops active queries from piling up.
- `idle_in_transaction_session_timeout = 10s`: kills transactions left open and
  idle. We basically never hit this (max 2 in 30 days), so it's just a safety net.

Both are config driven (`PG_*_MS`, `0` = no limit) and re-applied on every
reconnect, so you can tune them per env without touching code.

Connections also set `application_name="mlpa:{db}"` now, so MLPA shows up in the
CloudSQL `num_backends_by_application` metric and we can finally see how many
connections are ours.

## Ticket: 
[AIPLAT-921](https://mozilla-hub.atlassian.net/browse/AIPLAT-921)
PR companion: https://github.com/mozilla/dataservices-infra/pull/2088


src/mlpa/core/pg_services/pg_service.py
```diff
@@ -10,9 +11,14 @@
...
     pg: asyncpg.Pool | None
 
-    def __init__(self, db_name: str):
+    def __init__(self, db_name: str, statement_timeout_ms: int | None = None):
         self.db_name = db_name
         self.db_url = f"{cast(str, env.PG_DB_URL).rstrip('/')}/{db_name}"
+        self.statement_timeout_ms = (
+            statement_timeout_ms
+            if statement_timeout_ms is not None
+            else env.PG_STATEMENT_TIMEOUT_MS
```
noahpodgurski, line 21: question: If we're defining it here, do we still need to do 
```
async with self.statement_timeout(
            env.PG_MAINTENANCE_STATEMENT_TIMEOUT_MS
        ) as conn:
```
on every query?

src/mlpa/core/pg_services/litellm_pg_service.py
```diff
@@ -106,16 +109,19 @@ async def count_users_by_service_type(self) -> dict:
         `{base_user_id}:{service_type}`.
         """
         try:
-            rows = await self.pool.fetch(
```
noahpodgurski, line 109: question: Are we sure we can safely remove all the pool methods?

subpath: Only the three heavy read queries (list_users, count_users_by_service_type, list_managed_base_identities) moved from autocommit self.pool.fetch(...) to async with self.statement_timeout(...) as conn: conn.fetch(...)

src/mlpa/core/pg_services/pg_service.py
subpath: The one in connect() is the default that applies to every query ( 3s). statement_timeout() doesn't redefine it, it temporarily raises it for the few queries that genuinely need longer (full-table admin reads, startup reconciliation). So it's only on 4 call sites I belive

noahpodgurski (approved):
LGTM, let's make sure to test in dev first :)

PR #185, opened 2026-06-22
feat(readiness-probe-improvements): AIPLAT-916

subpath:
Jira ticket: [AIPLAT-916](https://mozilla-hub.atlassian.net/browse/AIPLAT-916)
Infra PR companion should be merged first: https://github.com/mozilla/dataservices-infra/pull/2090

## What's New

/health/readiness now answers `"can this pod serve a request right now?"` via the HTTP status code, instead of always returning 200. 

Kubernetes routes on the status code, so a pod with a down Postgres pool or a stale app_attest schema now drains from rotation instead of serving 500s.

Three checks run concurrently (asyncio.gather), each bounded by READINESS_CHECK_TIMEOUT_S (2s). All pass -> 200; any fail -> 503 with the failing check named in the body:

1. litellm pool: real SELECT 1 (replaces the old _closed-flag read, which only knew if .close() was called).
2. LiteLLM readiness: requires HTTP 200 and db == "connected" (the old code parsed the body but never checked the status, so a 503-with-body read as healthy).

Liveness stays a constant {"status": "alive"}, so a dependency blip drains the pod without ever restarting it, no crash-loop path from this change.

## QA: 
all new unit tests ✅ 

### Local QA: 

Every real "can't serve" condition now returns a deliberate 503 degraded with a body that says which check failed, the failed checks don't mask each other, liveness stays 200 so there's no restart loop, and the pod self-heals to 200 once the dependency returns.
image


src/mlpa/core/routers/health/health.py
```diff
@@ -36,27 +46,95 @@ async def liveness_probe():
...
+        get_litellm_version(client),
+        return_exceptions=True,
     )
-    litellm_status = response.json()
-    current_litellm_version = await get_litellm_version(client)
 
-    return {
-        "status": "connected",
+    # litellm pool — ping() never raises, but guard the gather contract anyway.
+    postgres_connected = litellm_ok is True
```
noahpodgurski, line 92: nit: Any reason we're using `is True` instead of `== True` here?

src/mlpa/core/routers/health/health.py
```diff
@@ -36,27 +46,95 @@ async def liveness_probe():
...
+    except Exception:
+        logger.error(
+            "readiness: could not resolve expected Alembic heads", exc_info=True
+        )
+        expected = set()
+        heads_resolved = False
+
+    migration_ok = heads_resolved and app_attest_connected and current == expected
+
+    # get_litellm_version() swallows its own errors, but guard the gather contract.
```
noahpodgurski: question: What does this comment mean?

src/mlpa/core/migrations.py
```diff
@@ -0,0 +1,34 @@
...
+    # Resolved lazily (not at import) so a path problem fails readiness gracefully
+    # rather than crashing app startup.
+    paths = list(mlpa.__path__)
+    if not paths:
+        raise RuntimeError("mlpa.__path__ is empty; cannot locate alembic/")
+    return Path(paths[0]).resolve().parents[1] / "alembic"
+
+
+@lru_cache(maxsize=1)
+def expected_heads() -> frozenset[str]:
```
noahpodgurski: question: What does this function do and why do we need to cache it?

src/mlpa/core/routers/health/health.py
subpath: `get_litellm_version()` catches its own errors and never raises in practice, but it runs inside `gather(return_exceptions=True)`, where any slot can still come back as an `Exception` (e.g. cancellation). The next line coerces that case to `"N/A"` so `version` is always a string. "Guard the gather contract" = handle the exception-object case.

src/mlpa/core/migrations.py
subpath: it reads alembic/ head revisions that is currently shiped and compars against DB revisions. I'm caching because files are immutable at runtime and probes are frequent.

src/mlpa/core/routers/health/health.py
subpath: I think that `is True litellm_ok` is bool or Exception from gather
is True matches only the singleton

noahpodgurski (approved):

PR #186, opened 2026-06-23
feat: search-dev budget

noahpodgurski:
## What's new

- Add `search-dev` budget
https://mozilla-hub.atlassian.net/browse/AIPLAT-988


src/mlpa/run.py
```diff
@@ -233,7 +233,7 @@ async def search(
         AuthorizedSearchRequest, Depends(authorize_search_request)
     ],
 ):
-    if authorized_search_request.service_type != "search":
+    if authorized_search_request.service_type not in {"search", "search-dev"}:
```
subpath: nitpick: 
Just thinking out loud, we should move this tuple to the config

subpath (approved):

src/mlpa/run.py
noahpodgurski: Agreed 👍

PR #189, opened 2026-06-26
fix: stream-contextvar-crashregression

subpath:
## Context: 
#184 moved the loguru log-field binding inside the streaming generator so request fields (user/model/service_type) stay attached to mid-stream logs. It used logger.contextualize(**log_fields) wrapped around the SSE yield. That introduced a regression.

stream_completion is an async generator. When the client disconnects mid-stream, Starlette tears the generator down from a different asyncio Task than the one that drove it. contextualize sets a contextvars token on enter and runs context.reset(token) on exit. The reset then runs in a foreign Context and raises:

`ValueError: <Token ...loguru_context...>` was created in a different Context

Sentry started seeing this right after #184 merged: 17 events, all on the streaming path (stream_completion), in production. It's handled and 0 users affected (the client already left by the time teardown runs), so no broken responses, but it's a steady stream of "Task exception was never retrieved" noise and the contextvar never gets reset. The extra finally: await gen.aclose() #184 added also gives a second close path that can feed the older aclose(): already running error.

## Fix

Bind the fields with logger.bind(**log_fields) instead of logger.contextualize(...). bind returns a child logger with no token to reset, so teardown from any task is clean. This keeps #184's actual goal (fields stay attached during streaming) without the contextvar lifecycle.

- Collapsed the wrapper + inner generator back into one generator.
- Routed the in-generator log calls through the bound logger.
- Threaded the bound logger into raise_and_log via a new log= kwarg so mid-stream error logs keep the request fields.

get_completion and get_search are plain coroutines, not generators, so contextualize enter/exit stays in one context there. Left unchanged.

## QA

Added test_stream_completion_aclose_from_separate_task_does_not_crash: drives the generator one chunk, then closes it from a separate asyncio task (separate Context, what Starlette does on disconnect). Against the current code it raises the exact prod ValueError at context.reset(token); with the fix it closes cleanly and records one ABORT, zero ERROR.

### DIFF
--- a/src/mlpa/core/utils.py
+++ b/src/mlpa/core/utils.py
@@ -221,7 +221,7 @@ def raise_and_log(
     response_code: int | None = None,
     response_text_prefix: str | None = None,
     *,
-    log=...,
+    log=logger,
 ) -> bytes: ...
 
 
@@ -232,7 +232,7 @@ def raise_and_log(
     response_code: int | None = None,
     response_text_prefix: str | None = None,
     *,
-    log=...,
+    log=logger,
 ) -> NoReturn: ...
 
 
--- a/src/mlpa/core/utils.py
+++ b/src/mlpa/core/utils.py
@@ -220,6 +220,8 @@ def raise_and_log(
     stream: Literal[True],
     response_code: int | None = None,
     response_text_prefix: str | None = None,
+    *,
+    log=...,
 ) -> bytes: ...
 
 
@@ -229,6 +231,8 @@ def raise_and_log(
     stream: Literal[False] = False,
     response_code: int | None = None,
     response_text_prefix: str | None = None,
+    *,
+    log=...,
 ) -> NoReturn: ...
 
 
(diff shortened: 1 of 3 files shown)


src/mlpa/core/utils.py
```diff
@@ -229,6 +231,8 @@ def raise_and_log(
     stream: Literal[False] = False,
     response_code: int | None = None,
     response_text_prefix: str | None = None,
+    *,
+    log=...,
```
noahpodgurski: question: This should be `log=logger`?

src/mlpa/core/utils.py
```diff
@@ -220,6 +220,8 @@ def raise_and_log(
     stream: Literal[True],
     response_code: int | None = None,
     response_text_prefix: str | None = None,
+    *,
+    log=...,
```
noahpodgurski: same here

noahpodgurski (approved):
Looks fine but 1 question

src/mlpa/core/utils.py
subpath: It's a PEP convention i used to : ) 
I will make it explicit, thanks 👍

PR #193, opened 2026-06-30
feat: answer service-type

noahpodgurski:
## What's new

- Add `answer` service type for exa `/answers` completion via `/v1/chat/completions`
- Refactored search service type check in `run.py` to reuse `valid_service_type_for_model` function
- Update `start.sh` to also run `mlpa` at the end of the script
https://mozilla-hub.atlassian.net/browse/AIPLAT-982


subpath (approved):

src/mlpa/run.py
```diff
@@ -29,6 +29,7 @@
 from mlpa.core.metrics import (
     SEARCH_MODEL,
     record_chat_availability,
+    record_chat_availability_for,
```
randy-concepcion, line 32: Good call on removing the call to `record_chat_availability_for`.  This can be removed too since it's no longer used.

PR #198, opened 2026-07-07
fix: MLPA disconnect causes Runtime Error

FieldMarshallObvious:
## What's Changed
- Removes manual await `response.aclose()` in cancel branch
- Suppresses runtime errors in httpx teardown
- Added tracking to capture dict to allow mocks to track how many times the stream was closed
- Added tests to verify that stream close call and validate that errors are not thrown

## QA Path
Added two test cases, added a test case to ensure that the stream is called once per close and that Runtime error is not thrown. 

### Close Count
In this test we run the close count validation on the old logic versus the new logic

Old Logic:
image

New Logic:
image

### Runtime error
In this test we validate that runtime errors are suppressed in the httpx teardown

Old Logic:
image

New Logic:
image

## Related Tickets & Documents
[AIPLAT-1124](https://mozilla-hub.atlassian.net/browse/AIPLAT-1124?atlOrigin=eyJpIjoiNTFiZGJlN2YzN2EzNGY2NGI3NDU3MjU5YmE2ZjVmZjIiLCJwIjoiaiJ9)

### DIFF
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -190,6 +190,7 @@ async def stream_completion(
                         asyncio.CancelledError,
                         StopAsyncIteration,
                         httpx.ReadError,
+                        RuntimeError,
                     ):
                         await next_chunk_task
                     break
--- a/src/mlpa/core/completions.py
+++ b/src/mlpa/core/completions.py
@@ -192,8 +192,6 @@ async def stream_completion(
                         httpx.ReadError,
                     ):
                         await next_chunk_task
-                    with contextlib.suppress(httpx.ReadError):
-                        await response.aclose()
                     break
 
                 try:
@@ -305,6 +303,7 @@ async def stream_completion(
                 asyncio.CancelledError,
                 StopAsyncIteration,
                 httpx.ReadError,
+                RuntimeError,
             ):
                 await next_chunk_task
         # Cancel the disconnect watcher and wait for it to finish to avoid
--- a/src/tests/unit/test_completions.py
+++ b/src/tests/unit/test_completions.py
@@ -1,6 +1,7 @@
 import asyncio
 import contextlib
 import json
+from typing import Callable
 from unittest.mock import AsyncMock, MagicMock
 
 import httpx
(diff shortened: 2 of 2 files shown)


src/mlpa/core/completions.py
randy-concepcion, line 189: `thought`: To add to the "belt-and-suspenders" part of the ticket, I'm wondering if adding the `RuntimeError` parameter should be added here too.

randy-concepcion (approved):

subpath:
I suggest to approach it with TDD : ) 


First make a test to reproduce the same error from the ticket, it should be failing - that would be RED
then you make a fix in completions.py and the test from step 1 should start passing
then refactor if needed

FieldMarshallObvious:
> I suggest to approach it with TDD : )
> 
> First make a test to reproduce the same error from the ticket, it should be failing - that would be RED then you make a fix in completions.py and the test from step 1 should start passing then refactor if needed

Added the tests you suggested and updated the PR message!

subpath:
> > I suggest to approach it with TDD : )
> > First make a test to reproduce the same error from the ticket, it should be failing - that would be RED then you make a fix in completions.py and the test from step 1 should start passing then refactor if needed
> 
> Added the tests you suggested and updated the PR message!

Thank you ❤️

src/mlpa/core/completions.py
subpath, line 189: suggestion:
it should be this, because cancelling this task can also raise RuntimeError, and without suppressing it here, disconnects turn into fake 502s instead of clean aborts

```python
with contextlib.suppress(
    asyncio.CancelledError,
    StopAsyncIteration,
    httpx.ReadError,
    RuntimeError,
):
    await next_chunk_task
```

subpath (approved):
https://github.com/Firefox-AI/MLPA/pull/198#discussion_r3549735733

after than it's good to go, thank you ❤️

src/mlpa/core/completions.py
FieldMarshallObvious: Sounds good! Added it in commit b0d5d4527605247f141c77c74c1518f421336fb3

FieldMarshallObvious: Added in commit b0d5d4527605247f141c77c74c1518f421336fb3!

PR #200, opened 2026-07-13
feat: make httpx/asyncpg debug logging lazy

noahpodgurski:
## What's new

  - Make HTTPX request/response debug logs lazy with `logger.opt(lazy=True)`.
  - Make asyncpg execute debug logs lazy for `query`, `args`, and `result`.
  - Move HTTPX JSON body redaction/truncation behind the lazy DEBUG log path.
  - Keep HTTPX/asyncpg error logs unconditional, so transport/DB failures still log at
  INFO+.

When `LOGURU_LEVEL=INFO`, DEBUG log payloads are no longer formatted before
  being dropped. This saves a small amount of CPU and avoids formatting request args/
  user IDs unless DEBUG logging is intentionally enabled.
https://mozilla-hub.atlassian.net/browse/AIPLAT-1143

### DIFF
--- a/src/mlpa/core/logger.py
+++ b/src/mlpa/core/logger.py
@@ -102,30 +102,36 @@ def _truncate_mapping(mapping, limit=5):
     return result
 
 
+def _httpx_params_repr(params):
+    return _truncate_mapping(params) if isinstance(params, dict) else _truncate(params)
+
+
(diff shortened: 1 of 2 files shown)


randy-concepcion (approved):
LGTM!

PR #202, opened 2026-07-13
feat: clamp prometheus metrics values to save memory buildup

noahpodgurski:
## What's new

- Add clamping to the following to keep prometheus metrics tracking low memory and quick response time on `/metrics`:
  - service type
  - purpose
  - model
  - country
  - request method
  - endpoint
- Removed `tool_name` from metrics (we can also add an allowlist for this if we want)
- Optimization: Keep ordered service-type lists for docs/errors, while using sets for O(1) metric label lookup
- Optimization: Refactored `clamp_country` to use a predefined frozen_set rather than regex

*Before*:
```
  clamp_country(valid,regex)     0.073 us
```

After:
```
  clamp_country(valid,set)       0.017 us
```

Total:
```
  clamp_model(valid)               0.035 us/call
  clamp_model(invalid)             0.033 us/call
  clamp_service_type(valid)        0.026 us/call
  clamp_service_type(invalid)      0.026 us/call
  clamp_purpose(valid)             0.026 us/call
  clamp_purpose(invalid)           0.027 us/call
  clamp_country(valid)             0.018 us/call
  clamp_country(invalid)           0.016 us/call
  clamp_country(none)              0.016 us/call
  clamp_request_method(valid)      0.032 us/call
  clamp_request_method(invalid)    0.029 us/call
  clamp_endpoint(matched)          0.032 us/call
  clamp_endpoint(unmatched)        0.025 us/call
  all clamps typical               0.176 us/call
```

https://mozilla-hub.atlassian.net/browse/AIPLAT-1142


src/mlpa/core/config.py
```diff
@@ -268,6 +295,29 @@ def valid_service_type_for_model(self, service_type: str, model: str) -> bool:
...
 
+    @cached_property
+    def valid_model_labels(self) -> set[str]:
+        """
+        Valid model label values for Prometheus.
+
+        Keep this in sync with the LiteLLM model list/search tools. It is defined
+        here rather than parsed from LiteLLM config because the MLPA pod should
+        not depend on that file existing at runtime.
+        """
```
randy-concepcion, line 306: `praise`: Good call on the comment here. Drift was my first thought and this seems to be the right call.

src/mlpa/core/metrics.py
```diff
@@ -113,23 +124,17 @@ def extract_tool_names(items: Iterable[dict]) -> list[str]:
 def record_request_with_tools(req: AuthorizedChatRequest) -> None:
     if not req.tools:
         return
-    for name in extract_tool_names(req.tools):
-        metrics.chat_requests_with_tools.labels(
-            tool_name=name, **_chat_labels(req)
-        ).inc()
+    metrics.chat_requests_with_tools.labels(**_chat_labels(req)).inc()
```
randy-concepcion: `question`: I'm not sure if the model is clamped here yet if this runs before the request goes to LiteLLM. Should `_chat_labels()`  here be called with `bounded_model=True` similar to the other  `_chat_labels()` calls?

randy-concepcion (commented):
Looks great overall! 🙌  Just had one question to confirm the clamping behavior of model names getting passed in one of the methods.

src/mlpa/core/metrics.py
noahpodgurski: Yes you're right, good callout! 🙌

src/mlpa/core/config.py
noahpodgurski: Yes, of course we just have to be on top of it

randy-concepcion (approved):
Nice, LGTM! 🚀

PR #212, opened 2026-07-23
feat: add release candidate tagging workflow

randy-concepcion:
## What's New

- Adds `.github/workflows/tag-release-candidate.yml`, a `workflow_dispatch` workflow that tags an existing main-build image in GAR as a release candidate (`vX.Y.Z-rc.N`)
- Tags the existing digest directly (no pull, no rebuild) via `gcloud artifacts docker tags add`
- Inputs: a full commit SHA and an RC number
- Validates the commit is on `main`, reads the version from `pyproject.toml`, and derives the RC tag
- Safe to re-run: re-tagging the same digest is a no-op, and it refuses to move an existing RC tag to a different digest. Fails closed if the source image is missing or the registry lookup errors.
- Keyless auth via the existing `artifact-writer` Workload Identity setup, no new secrets or permissions

## Context

This PR is part of [AIPLAT-911](https://mozilla-hub.atlassian.net/browse/AIPLAT-911) (per-environment image tiers). It establishes the stage tier for MLPA: dev runs `latest`, stage will pin to an RC tag, and prod stays on stable semver.

The workflow promotes a selected main-build image to `vX.Y.Z-rc.N` by adding a tag to the existing digest in GAR. It does not rebuild, pull, or push any image. A separate follow-up PR will add a stable promotion workflow triggered on GitHub release.

Every step after the build tags the same digest.
```mermaid
flowchart LR
    A[merge to main] --> B["build<br/>(short-SHA tag)"]
    B --> C["this PR<br/>adds vX.Y.Z-rc.N"]
    C --> D["validate RC on stage<br/>(manual)"]
    D --> E["create GitHub release<br/>(manual)"]
    E --> F["follow-up PR<br/>adds vX.Y.Z"]

    style C fill:#e8f4f8
```

## Testing

Not yet run. A `workflow_dispatch` workflow only appears in the Actions "Run workflow" UI after it lands on the default branch, so it can't be triggered from the PR branch. Will validate post-merge by promoting a known `main` build (latest) and confirming the RC tag lands on the expected digest.

## How to use

From the Actions tab, select "Tag Release Candidate in GAR", click "Run workflow" on `main`, and provide:
- The full commit SHA of a main-branch build
- The RC number (1, 2, 3, ...)

The workflow produces a job summary table with the commit, source tag, version, RC tag, and digest.

## Related

- [AIPLAT-911](https://mozilla-hub.atlassian.net/browse/AIPLAT-911)

PR #213, opened 2026-07-23
feat: add smoke tests and post-deploy workflow

noahpodgurski:
## What's new

- Updated `linters_and_tests.yml` to separate out the docs verification into separate steps and for running smoke tests.
- Implemented smoke tests for App Attest QA, Play Integrity, FxA, SmartWindow, and Memories data flows. Each smoke test tests by local function invoking and remote deployments (Play Integrity testing path not yet finalized).
- Updated dependencies in pyproject.toml and uv.lock to use the `v0.9.2` of [PyFxA-Oxide](https://github.com/mozilla/PyFxA-Oxide/releases/tag/v0.9.2) with `CI_WAF_TOKEN` changes, required for FxA testing.
- Refactored app_attest_qa.py to remove unnecessary imports and improve payload handling.
- Removed `mock_response` from `classes`, since LiteLLM doesn't seem to support that body parameter anymore.
- Added `mock_response` as a separate model `mock` for testing and adjusted test assertions for consistency.
https://mozilla-hub.atlassian.net/browse/AIPLAT-908


.github/workflows/post-deploy-smoke.yml
```diff
@@ -0,0 +1,165 @@
...
+        type: string
+      blocking:
+        required: false
+        default: false
+        type: boolean
+    secrets:
+      # ARGOCD_AUTH_TOKEN:
+      #   required: true
+      # ARGOCD_SERVER:
+      #   required: true
```
noahpodgurski, line 45: I can add a TODO: here, or we can create another ticket for this. The idea is to hook the test job up to argo and have it check the sync status, and then fire off the remote smoke tests

.github/workflows/post-deploy-smoke.yml
```diff
@@ -0,0 +1,165 @@
...
+      #     ARGOCD_AUTH_TOKEN: ${{ secrets.ARGOCD_AUTH_TOKEN }}
+      #     ARGOCD_SERVER: ${{ secrets.ARGOCD_SERVER }}
+      #   run: |
+      #     argocd app wait "${{ inputs.argo_app }}" \
+      #       --server "$ARGOCD_SERVER" \
+      #       --auth-token "$ARGOCD_AUTH_TOKEN" \
+      #       --grpc-web \
+      #       --sync \
+      #       --health \
+      #       --timeout 900
```
noahpodgurski, line 109: See my comment above, this is some idea of how that integration might look

.github/workflows/post-deploy-smoke.yml
```diff
@@ -0,0 +1,165 @@
...
+        continue-on-error: true
+        env:
+          SMOKE_APP_BUNDLE_ID: ${{ vars.SMOKE_APP_BUNDLE_ID || 'org.mozilla.ios.Fennec' }}
+          SMOKE_APP_DEVELOPMENT_TEAM: ${{ secrets.SMOKE_APP_DEVELOPMENT_TEAM }}
+          SMOKE_BASE_URL: ${{ inputs.base_url }}
+          SMOKE_FXA_TOKEN: ${{ secrets.SMOKE_FXA_TOKEN }}
+          SMOKE_PLAY_INTEGRITY_TOKEN: ${{ secrets.SMOKE_PLAY_INTEGRITY_TOKEN }}
+          SMOKE_PLAY_INTEGRITY_USER_ID: ${{ secrets.SMOKE_PLAY_INTEGRITY_USER_ID }}
+          CI_WAF_TOKEN: ${{ secrets.CI_WAF_TOKEN }}
+        run: uv run pytest src/tests/smoke -v
```
noahpodgurski, line 134: The smoke tests are ran in two forms essentially. It tests against the `mocked_client_integration` if `SMOKE_BASE_URL` is not set. If it is set, then it runs the tests against that deployment. App Attest, FxA, and Play Integrity auth are required. I've hooked up a test path for all except play integrity so far.

src/mlpa/core/classes.py
```diff
@@ -16,7 +16,6 @@ class ChatRequest(BaseModel):
     temperature: Optional[float] = env.TEMPERATURE
     max_completion_tokens: Optional[int] = env.MAX_COMPLETION_TOKENS
     top_p: Optional[float] = env.TOP_P
-    mock_response: Optional[str] = None
```
noahpodgurski, line 19: In my testing, I couldn't get this parameter to work anymore, maybe it was an uptick in LiteLLM version that made it deprecated, but I only found success through setting a dedicated mock model through the litellm config
```
model_list:
  - model_name: mock
    litellm_params:
      model: openai/gpt-4o
      api_key: "mock-token"
      mock_response: "this is a mocked response"
```

src/tests/integration/conftest.py
```diff
@@ -1,108 +1,4 @@
...
-    MockAppAttestPGService,
-    MockFxAClientForMockRouter,
-    MockFxAService,
-    MockLiteLLMPGService,
-    mock_app_attest_auth,
-    mock_get_completion,
-    mock_get_or_create_user,
-    mock_verify_assert,
-    mock_verify_attest,
+from tests.fixtures import (  # noqa: F401
```
noahpodgurski, line 1: These imports are required for the test fixtures, and the comment tells the linter to not complain about the unused imports

src/tests/smoke/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+
+from tests.consts import TEST_FXA_TOKEN
+from tests.fixtures import (  # noqa: F401
+    mocked_client_integration,
+    use_real_get_or_create_user,
+)
+
+FXA_PASSWORD = "123dev123dev123dev"
+FXA_CLIENT_ID = "5882386c6d801776"
```
noahpodgurski, line 15: This is our public stage FxA client ID which we use in a few places throughout MLPA

src/tests/smoke/test_auth_data_flows.py
```diff
@@ -0,0 +1,434 @@
...
+    "key_id.json",
+    "root_cert.pem",
+    "root_key.pem",
+)
+DEFAULT_APP_ATTEST_QA_BUCKET = f"gs://{env.APP_ATTEST_QA_BUCKET}/{(env.APP_ATTEST_QA_BUCKET_PREFIX or '').strip('/')}"
+
+
+def _chat_payload() -> dict:
+    return {
+        "model": MOCK_MODEL_NAME,
```
noahpodgurski, line 38: `MOCK_MODEL_NAME` resolves to `mock`, which should be a configured mock model on the `remote` we're testing against (litellm configuration mentioned in comment above)

.github/workflows/post-deploy-smoke.yml
```diff
@@ -0,0 +1,165 @@
...
+        env:
+          SMOKE_APP_ATTEST_QA_CERT_B64: ${{ secrets.SMOKE_APP_ATTEST_QA_CERT_B64 }}
+        run: |
+          if [ -z "$SMOKE_APP_ATTEST_QA_CERT_B64" ]; then
+            echo "No App Attest QA certificate bundle configured; live App Attest smoke will skip."
+            exit 0
+          fi
+          mkdir -p "$RUNNER_TEMP/app-attest-qa"
+          printf '%s' "$SMOKE_APP_ATTEST_QA_CERT_B64" | base64 -d | tar -xz -C "$RUNNER_TEMP/app-attest-qa"
+          echo "APP_ATTEST_QA_CERT_DIR=$RUNNER_TEMP/app-attest-qa" >> "$GITHUB_ENV"
```
noahpodgurski, line 121: This decodes the base64 encoded tar.gz cert bundle from `SMOKE_APP_ATTEST_QA_CERT_B64`,extracts the qa certs, and exports that dir in `APP_ATTEST_QA_CERT_DIR` for the smoke tests

.github/workflows/linters_and_tests.yml
```diff
@@ -42,8 +42,56 @@ jobs:
...
+        run: pip install uv==0.10.8
+
+      - name: Sync dependencies
+        run: |
+          uv venv
+          uv sync --all-groups
+          uv pip install --no-cache-dir -e .
+
+      - name: Run smoke tests
+        run: uv run pytest src/tests/smoke -v
```
randy-concepcion, line 97: `question`: I think the ticket mentioned to also add the existing `src/tests/e2e` tests too. Just curious, is that too complex to be added to the pipeline? It's a good test, but can see it taking a lot to standup for the test.

docs/smoke-tests.md
```diff
@@ -0,0 +1,42 @@
...
+- App Attest QA path, including `use-qa-certificates: true`
+- Play Integrity
+- FxA
+- SmartWindow chat data flow
+- Memories data flow
+
+These tests assert HTTP status and OpenAI-compatible response shape. They do not
+assert model text. Local smoke uses a completion mock. Post-deploy
+smoke runs the same request shape against the deployed `/v1/chat/completions`
+path via mocked mock `mock`.
```
randy-concepcion, line 15: `question`/nit`: Is this referring to the mock model `mock` in the `litellm_config.yaml` file?

.github/workflows/linters_and_tests.yml
```diff
@@ -42,8 +42,56 @@ jobs:
...
+        run: |
+          uv venv
+          uv sync --all-groups
+          uv pip install --no-cache-dir -e .
+
       - name: Build docs
         run: make docs
 
       - name: Verify docs are up to date
         run: git diff --exit-code -- docs/index.html
```
randy-concepcion, line 74: `thought`: Unrelated to your changes, but just a heads up that these will fail due to a recently released version update to `styled-components` package that `redocly` uses. I fixed that in [this commit,](https://github.com/Firefox-AI/MLPA/pull/212/changes/ce27809114bd55a41591f993bc4569c2a215f915) but unsure how you'd like to handle that (i.e., take the change in your PR, merge my PR first, or split into its own PR...it's a small change).

src/tests/unit/test_app_attest_qa_script.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    decoded = cbor2.loads(assertion)
+    assert set(decoded) == {"authenticatorData", "signature"}
+    assert isinstance(decoded["authenticatorData"], bytes)
+    assert isinstance(decoded["signature"], bytes)
+    assert int.from_bytes(decoded["authenticatorData"][33:37], "big") == 1
+
+
+def test_compute_payload_hash_matches_httpx_json_encoding():
```
randy-concepcion, line 57: `praise`: Nice job! 🙌 

`suggestion`: Could we add a non-ASCII version of this test, e.g., payload with an emoji? The `ensure_ascii=False` fix is solid, but having this test could protect against any regressions.

.github/workflows/linters_and_tests.yml
```diff
@@ -42,8 +42,56 @@ jobs:
       - name: Run integration tests
         run: uv run pytest src/tests/integration -v
 
+  build-and-verify-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
```
randy-concepcion, line 51: `thought`: No action needed here, but I noticed the python/uv/sync/install config blocks in a few places. We may want to consider in the future putting them into a small composite action so its reusable and we can avoid potential version/config drift. 👍

.github/workflows/post-deploy-smoke.yml
randy-concepcion: Thanks for clarifying this. I think the work here is substantial enough for creating a new ticket and adding the link in a TODO here. Have we integrated Argo in our workflows before? I don't remember seeing anything here, so there's some level of uncertainty that could also lean more in favor of creating a ticket for this.

.github/workflows/post-deploy-smoke.yml
randy-concepcion: Oh nice! You already have config here. Same as I mentioned above, this should prob be split off in its own ticket. 👍 I'm curious to see how this works!

src/mlpa/core/classes.py
randy-concepcion: `thought`: Thanks for clarifying this. I think removing this makes sense. 👍

src/tests/smoke/test_auth_data_flows.py
```diff
@@ -0,0 +1,434 @@
+import base64
```
randy-concepcion, line 1: `praise`: Great work 🔥

randy-concepcion (commented):
Awesome work getting the smoke test setup + initial tests in place! I just had a few comments and thoughts, mostly around splitting some of the work into follow-up tickets (e.g., e2e test wiring, Argo/post-deploy hookup). I also had one suggestion for adding a non-ASCII test to round out the encoding coverage. Other than that, this looks great! 🙌

.github/workflows/post-deploy-smoke.yml
noahpodgurski: Agreed, it'd be new territory

.github/workflows/linters_and_tests.yml
noahpodgurski: We have a separate `e2e` tests folder, which I think at this point should be deprecated/renamed in favor of this `smoke` tests folder, which is more comprehensive especially when doing a remote test. I also forgot to mention the remote test can be manually triggered in github actions and after testing (and the argo sync stuff) we can make it automatic, which should encapsulate a more e2e flow

docs/smoke-tests.md
noahpodgurski: Yes, we'll also want to have the same configuration in dev, or environments where we want to test against

.github/workflows/linters_and_tests.yml
noahpodgurski: We can merge yours first 👍

src/tests/unit/test_app_attest_qa_script.py
noahpodgurski: Sure, good callout

.github/workflows/linters_and_tests.yml
noahpodgurski: That's a great cleanup/refactor ticket idea! 🙌

subpath:
@noahpodgurski  this is a great work ❤️ 
Maybe we could QA the post-deploy workflow in this PR? 🙏 

Also praise @randy-concepcion  detailed feedback ❤️

noahpodgurski:
> @noahpodgurski this is a great work ❤️ Maybe we could QA the post-deploy workflow in this PR? 🙏
> 
> Also praise @randy-concepcion detailed feedback ❤️

I think we have to merge to main first to run the workflows right? I've QA'd it locally against dev though

subpath (approved):

PR #217, opened 2026-07-28
feat: basic fuzzing protection for model name and ios, android tokens

subpath:
## Description
Reject fuzzing/injection payloads on model, App Attest key_id_b64, and Play Integrity integrity_token before they cost a DB round trip or an external API call.

We were seeing scanner traffic (SQLi, XSS, code injection) landing in these free-text fields. None of it can actually do anything to this stack, but it still triggers a DB query or an external call (LiteLLM, Google Play Integrity) before getting rejected downstream.

Added plain charset + length checks, no whitelist to maintain:
- model: lowercase ascii + digits + .-, max 64 chars, checked in authorize_chat_request before any auth/DB/LiteLLM work
- key_id_b64 (App Attest challenge): base64 charset, max 128 chars, checked before the DB lookup
- integrity_token (Play Integrity): JWT-like charset (-_. + alnum), max 10000 chars, checked before the Google API call

All garbage still gets a 400, just without the round trip.

## QA
Unit tests table-driven over real model names / real-shaped tokens (should pass) and the sampled attack payloads (should reject), plus edge cases (empty, over length, bad edge chars). Integration tests confirm the downstream call (_decode_integrity_token, env.valid_service_type_for_model, fxa_auth) is never reached for rejected input, not just that the response is a 400.


src/mlpa/core/utils.py
```diff
@@ -39,6 +40,46 @@ def clamp_model(model: str) -> str:
...
+MAX_MODEL_NAME_LEN = 64
+
+
+def is_valid_model_name(model: str) -> bool:
+    if not model or len(model) > MAX_MODEL_NAME_LEN:
+        return False
+    if (
+        model[0] not in _ALLOWED_MODEL_EDGE_CHARS
+        or model[-1] not in _ALLOWED_MODEL_EDGE_CHARS
+    ):
```
noahpodgurski: question: This is an interesting approach. Wouldn't it be simpler to use a regex check?

src/mlpa/core/utils.py
```diff
@@ -39,6 +40,46 @@ def clamp_model(model: str) -> str:
...
+
+def is_plausible_base64_key_id(key_id_b64: str) -> bool:
+    if not key_id_b64 or len(key_id_b64) > MAX_KEY_ID_B64_LEN:
+        return False
+    return set(key_id_b64) <= _BASE64_STANDARD_CHARS
+
+
+# Play Integrity tokens are compact JWT/JWE strings (base64url segments joined
+# by "."); this bound is generous slack for the real token size, not exact.
+_JWT_LIKE_CHARS = frozenset(string.ascii_letters + string.digits + "-_.")
```
noahpodgurski, line 67: question: Can't base64 also include `+=/`?

src/mlpa/core/auth/authorize.py
```diff
@@ -120,6 +124,18 @@ async def authorize_chat_request(
...
+    if not is_valid_model_name(chat_request.model):
+        record_chat_availability_for(
+            AvailabilityReason.INVALID_MODEL_NAME,
+            model=chat_request.model,
+            service_type=service_type.value,
+            purpose=purpose or "",
+        )
+        raise HTTPException(
+            status_code=400,
+            detail={"error": ERROR_CODE_INVALID_MODEL_NAME},
```
noahpodgurski, line 138: thought: This makes sense to have a pre req model name check (and to also keep the current LiteLLM check), but maybe we could add some comments detailing the flow and purpose of each check?

src/mlpa/core/auth/authorize.py
```diff
@@ -12,11 +12,15 @@
     SearchRequest,
     ServiceType,
 )
-from mlpa.core.config import env
```
noahpodgurski, line 15: thought: Could we record some testing of how much latency these checks add to the request?

src/mlpa/core/utils.py
subpath: I think regex is harder to maintain 😬 
Easier to make mistakes

src/mlpa/core/utils.py
subpath: _BASE64_STANDARD_CHARS includes +/=

src/mlpa/core/auth/authorize.py
subpath: I would say microseconds

src/mlpa/core/utils.py
noahpodgurski: It's more efficient and comprehensive for what we're trying to check here though no? This currently is only checking the first and last characters of the model name, ideally each character should be verified against a regex like 


```
VALID_MODEL_NAME_PATTERN = re.compile(r"^[a-z0-9](?:[a-z0-9._/-]{0,62}[a-z0-9])?$")

def is_valid_model_name(model: str) -> bool:
    return bool(MODEL_NAME_PATTERN.match(model))
```

noahpodgurski: Actually, I think technically the regex check is slower, but it does cover more of the requirement

subpath:
```
set-based : 0.1825 us/call
regex     : 0.1672 us/call
regex is 0.92x the set-based latency (~15ns faster per call)
```

actually regex is a bit faster : ) 
Let's use regex, I agree

noahpodgurski (approved):

PR #221, opened 2026-08-04
feat: automatically tag main builds as release candidates

randy-concepcion:
## What's New

- Adds a `workflow_run` trigger to `tag-release-candidate.yml` so every successful main build for an unreleased version gets an RC tag automatically
- `workflow_dispatch` stays as the manual path, for re-tagging an older commit, recovering from a failed automatic run, or supplying a specific RC number
- `rc_number` is now optional. Left empty, the workflow allocates the next number for the current version.
- Adds `queue: max` to both promotion workflows so rapid merges queue instead of canceling each other's pending runs.
  - NOTE: Both need it since they share a concurrency group, and mismatched settings would behave differently depending on which queued first.

## Context

This ticket is for [AIPLAT-1188](https://mozilla-hub.atlassian.net/browse/AIPLAT-1188) (a follow-up to [AIPLAT-911](https://mozilla-hub.atlassian.net/browse/AIPLAT-911)). Tagging a release candidate was a manual workflow run. Since most merges to main end up shipping, and a main build is left with only its commit hash once `latest` moves to the next build, RC tagging is now automatic.

The workflow chains off the build completing rather than the push itself, since the image doesn't exist until the build finishes.

## Behavior

| Trigger | Condition | Result |
|---|---|---|
| Automatic | Version not yet released | Creates `vX.Y.Z-rc.N` where N is the highest existing + 1 |
| Automatic | Stable tag points at another commit | Skips cleanly, no tag created |
| Automatic | Digest already has an RC | No-op |
| Manual | No `rc_number` given | Allocates the next number |
| Manual | `rc_number` already on this digest | No-op |
| Manual | `rc_number` differs from the RC already on this digest | Fails, names the existing tag |
| Manual | Stable tag points at another commit | Fails |
| Either | Stable tag points at this commit | Tags anyway. The release was cut mid-run and promotion needs the RC tag. |

RC numbers are an allocation counter, not an ordering signal, since builds can finish out of commit order. The run summary is the authoritative tag-to-commit mapping.

## Testing

Can't be tested end to end before merge. `workflow_run` only fires from the default branch, and `workflow_dispatch` doesn't appear in the Actions UI until the file lands there. First validation is the next merge to main after this ships.

What was verified locally: the `tags list` projection against production GAR, the decision rule across all its branches, the stable-tag position logic against real lightweight and annotated tags, plus actionlint and shellcheck.

The manual path is unchanged, so if the automatic trigger misbehaves we can still tag a release candidate by running the workflow directly.

## Notes

**The first automatic run will skip.**
Main is at `1.0.24` and `v1.0.24` is already released on `7a1b1ea`, so every commit currently on main is in the post-release version gap. The first real RC tag lands on the `1.0.25` bump commit. This is expected.

**A release cut before its build finishes can make `promote-release` fail once.**
Both workflows share a FIFO queue, so promotion can run ahead of the RC job, fail its RC gate, and then the RC job creates the tag. Re-running the promote workflow resolves it once the RC tag exists.

**Don't add required reviewers to the `build` environment.**
It would block every automatic run on every merge to main.

### DIFF
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -6,9 +6,9 @@ on:
       - released
 
 concurrency:
-  # Shared with the RC tagging workflow so the two cannot mutate tags at the same time
+  # Serialize RC tagging and stable promotion without dropping pending runs.
   group: mlpa-image-promotion
-  cancel-in-progress: false
+  queue: max
 
 env:
   GAR_LOCATION: us
@@ -178,8 +178,8 @@ jobs:
           fi
 
           # Require a release candidate tag for this version on the source digest.
-          # This establishes continuity with the RC tier, but does not prove that
-          # stage testing passed.
+          # With automatic RC tagging, this confirms that the artifact reached the RC tier.
+          # It does not prove that it was deployed to stage or that stage validation passed.
           RC_PREFIX="${STABLE_TAG}-rc."
           matched_rc=""
           while IFS= read -r t; do
--- a/.github/workflows/tag-release-candidate.yml
+++ b/.github/workflows/tag-release-candidate.yml
@@ -1,6 +1,15 @@
 name: Tag Release Candidate in GAR
 
 on:
+  # Automatic: every successful main build for an unreleased version gets an
+  # RC tag, so the image keeps a durable semantic label after `latest` moves on.
+  # Chained off the build rather than the push itself, because the image does
+  # not exist until the build lands.
+  workflow_run:
+    workflows: ["Build and Push Docker Image to Google Artifact Registry"]
+    types: [completed]
+
+  # Manual: tag a specific commit, optionally with an explicit RC number.
   workflow_dispatch:
     inputs:
       commit_sha:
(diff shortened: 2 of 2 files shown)


noahpodgurski (approved):

FieldMarshallObvious (approved):
Looks good to me! 🚀

randy-concepcion:
**Scenarios validated**

* Automatic workflow w/ version already released  
  * Result → Skipped correctly, image in GAR remained intact (see [workflow run](https://github.com/Firefox-AI/MLPA/actions/runs/30938874949))  
* Automatic workflow w/ version not yet released  
  * Result → Created `v1.0.25-rc.1` (see [workflow run](https://github.com/Firefox-AI/MLPA/actions/runs/30940609102))  
* Manual workflow w/ no `rc_number` provided \+ digest already tagged  
  * Result → Failed (as expected) and named the existing tag (see [workflow run](https://github.com/Firefox-AI/MLPA/actions/runs/30941927617))  
* Manual workflow w/ `rc_number = 1` \+ digest already tagged  
  * Result → No-op/no change made (see [workflow run](https://github.com/Firefox-AI/MLPA/actions/runs/30942120707))

[**MLPA Image on GAR**](https://console.cloud.google.com/artifacts/docker/moz-fx-llm-proxy-prod-b0b8/us/llm-proxy-prod/mlpa/sha256:6d6a3e8657286e6192c6b388806e8bebeb6b1a4f3759da5a61489a653db95b8d?project=moz-fx-llm-proxy-prod-b0b8) (same digest, auto-tagged RC version after merge to main)



image

PR #223, opened 2026-08-04
feat: track firefox client version from user agent

noahpodgurski:
## What's new

- Add tracking of major firefox version based off `user-agent` header
- Parses major version with regex, defaulting to `""` or `"unknown"` if bad value
- Whitelists firefox version to 100-200, we can pull in the few latest versions from https://product-details.mozilla.org/1.0/firefox_history_stability_releases.json but that adds a dependency, so we're hard coding it for now
https://mozilla-hub.atlassian.net/browse/AIPLAT-1215

### DIFF
--- a/src/mlpa/core/middleware/instrumentation.py
+++ b/src/mlpa/core/middleware/instrumentation.py
@@ -1,4 +1,3 @@
-import re
 import time
 
 from fastapi import Request
@@ -10,9 +9,8 @@ from mlpa.core.utils import (
     clamp_purpose,
     clamp_request_method,
     clamp_service_type,
+    parse_firefox_major_version_from_user_agent,
 )
-
-FX_USER_AGENT_RE = r".*?Firefox/(\d+\.\d+)"
 
 
 async def instrument_requests_middleware(request: Request, call_next):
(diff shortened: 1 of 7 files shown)


src/mlpa/core/middleware/instrumentation.py
```diff
@@ -30,7 +38,22 @@ async def instrument_requests_middleware(request: Request, call_next):
...
             service_type = request.headers.get("service-type", "")
             purpose = request.headers.get("purpose", "")
-
+            user_agent = request.headers.get("user-agent", "")
+            major_fx_version = ""
+            try:
+                if user_agent:
+                    fx_user_agent_match = re.match(FX_USER_AGENT_RE, user_agent)
+                    fx_version = (
+                        fx_user_agent_match.group(1) if fx_user_agent_match else ""
```
FieldMarshallObvious: "" is doing a couple things here, should we seperate out the different branches? Like `none` - `non-firefox` - `no-match`

src/mlpa/core/middleware/instrumentation.py
```diff
@@ -30,7 +38,22 @@ async def instrument_requests_middleware(request: Request, call_next):
...
+                    )
+                    (major_fx_version, minor_fx_version) = (
+                        fx_version.split(".") if fx_version else ("", "")
```
FieldMarshallObvious: Should we simplify this to `major_fx_version = fx_version.split(".")[0]`? This would prevent silent errors in case of future regex edits.

src/mlpa/core/middleware/instrumentation.py
noahpodgurski: I think `""` meaning non-firefox is fine. `unknown` is a user-agent that doesn't whitelist to Firefox/ABC.XYZ

src/mlpa/core/middleware/instrumentation.py
noahpodgurski: Sure

randy-concepcion (approved):
Nice! LGTM! 🚀

PR #232, opened 2026-08-19
ci: sync GitHub releases to Jira versions + confluence notes

jgauf:


.github/scripts/release_to_jira.py
```diff
@@ -0,0 +1,237 @@
...
+# 6. Build release notes and publish as a child of the Release Notes Folder.
+def li(items):
+    return "".join(f"<li>{x}</li>" for x in items)
+
+
+def esc(s):
+    return html.escape(str(s), quote=True)
+
+
+ticket_html = (
```
noahpodgurski, line 155: thought: Not to overcomplicate what already is working here but there's python libraries out there that could make generating this HTML (and reading/understanding it) easier. For instance `airium`:

```python3
from airium import Airium

a = Airium()

with a.ul():
    if attached:
        for k in attached:
            with a.li():
                a.a(href=f"{SITE}/browse/{k}", _text=k)
    else:
        a.li(_text="None linked")

ticket_html = str(a)
```

noahpodgurski (approved):

.github/scripts/release_to_jira.py
subpath: I personally think it's fine we have only 4 small fragment here

subpath (approved):

PR #234, opened 2026-08-24
feat(country-metrics): AIPLAT-1266 per-country metrics

subpath:
We want country/region breakdown for chat metrics in Grafana.

Adds a bounded per-country dimension to chat completion latency, TTFT, and availability metrics. Only few markets (US, CA, FR, DE) get their own label, everything else buckets into "other" so we don't blow up cardinality (211 countries seen in prod today, see ).

Added a docs/prometheus-cardinality.md for future references

[AIPLAT-1266](https://mozilla-hub.atlassian.net/browse/AIPLAT-1266)

### DIFF
--- a/src/mlpa/core/classes.py
+++ b/src/mlpa/core/classes.py
@@ -87,18 +87,28 @@ class PlayIntegrityTokenResponse(BaseModel):
     expires_in: int
 
 
-class AuthorizedRequestLogMixin:
+class AuthorizedRequestLogMixin(BaseModel):
     """Shared structured log fields for authorized requests.
 
     Bound into the loguru contextvar via ``logger.contextualize(**log_fields)``
     in the proxy handlers so every log line emitted while serving the request
     (including mid-stream errors) carries them as queryable ``record.extra.*``
     fields, rather than concatenated into the message string.
+
+    A ``BaseModel`` itself (not a plain mixin) so ``ty`` resolves these fields
+    on ``AuthorizedChatRequest``/``AuthorizedSearchRequest`` through multiple
+    inheritance instead of flagging their constructor calls as unknown kwargs.
     """
 
     user: str
     service_type: str
-    purpose: str
+    purpose: str = (
+        ""  # From header; empty for service types without defined purposes (e.g. s2s)
+    )
+    # Raw X-Geo-Country, edge-stamped; clamp at metric time. Read by
+    # requests_by_country_total for both chat and search, and additionally by
+    # the latency/TTFT/availability by-country metrics (AIPLAT-1266) for chat.
+    client_country: str = ""
 
     @property
     def log_fields(self) -> dict[str, str]:
@@ -115,14 +125,7 @@ class AuthorizedRequestLogMixin:
 
 
 class AuthorizedChatRequest(ChatRequest, AuthorizedRequestLogMixin):
-    user: str
-    service_type: str
-    purpose: str = (
-        ""  # From header; empty for service types without defined purposes (e.g. s2s)
-    )
-    # Raw X-Geo-Country, edge-stamped; clamp at metric time. Read by the
-    # by-country chat metrics (AIPLAT-1266).
-    client_country: str = ""
+    pass
 
 
 class SearchRequest(BaseModel):
@@ -131,15 +134,7 @@ class SearchRequest(BaseModel):
 
 
 class AuthorizedSearchRequest(SearchRequest, AuthorizedRequestLogMixin):
-    user: str
-    service_type: str
-    purpose: str = (
-        ""  # From header; empty for service types without defined purposes (e.g. s2s)
-    )
-    # Raw X-Geo-Country, edge-stamped; clamp at metric time. Read by
-    # requests_by_country_total, but not by the latency/TTFT/availability
-    # by-country metrics (AIPLAT-1266), which are chat-only for now.
-    client_country: str = ""
+    pass
 
 
 # Dynamically create ServiceType enum from config
--- a/src/mlpa/core/classes.py
+++ b/src/mlpa/core/classes.py
@@ -120,7 +120,8 @@ class AuthorizedChatRequest(ChatRequest, AuthorizedRequestLogMixin):
     purpose: str = (
         ""  # From header; empty for service types without defined purposes (e.g. s2s)
     )
-    # Raw X-Geo-Country, edge-stamped; clamp at metric time.
+    # Raw X-Geo-Country, edge-stamped; clamp at metric time. Read by the
+    # by-country chat metrics (AIPLAT-1266).
     client_country: str = ""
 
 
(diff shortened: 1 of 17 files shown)


src/mlpa/core/classes.py
```diff
@@ -133,6 +135,10 @@ class AuthorizedSearchRequest(SearchRequest, AuthorizedRequestLogMixin):
     purpose: str = (
         ""  # From header; empty for service types without defined purposes (e.g. s2s)
     )
+    # Raw X-Geo-Country, edge-stamped; clamp at metric time. Only chat metrics
+    # (AIPLAT-1266) read this today; search carries it for parity but doesn't
+    # record a by-country metric yet.
+    client_country: str = ""
```
noahpodgurski: nit: I wonder if we could combine these definitions for `AuthorizedChatRequest` and `AuthorizedSearchRequest` somehow, and then reference them to minimize code duplication. Something like:
```
class CommonRequestHeaders(ChatRequest/SearchRequest, AuthorizedRequestLogMixin):
    user: str
    service_type: str
    purpose: str = ""
    client_country: str = ""

class AuthorizedChatRequest(ChatRequest, AuthorizedRequestLogMixin):
    (...CommonRequestHeaders)

class AuthorizedSearchRequest(SearchRequest, AuthorizedRequestLogMixin):
    (...CommonRequestHeaders)
```

If not that's fine, but can we at least align the comments above `client_country` to be the same on each line? 🙏

src/mlpa/core/config.py
```diff
@@ -18,6 +18,13 @@ class Env(BaseSettings):
     MLPA_CAPPED_SERVICE_TYPES: set[str] = {"ai", "memories"}
     MLPA_ADMISSION_LOCK_TIMEOUT_MS: int = 5000
 
+    # Countries with their own Grafana breakout (AIPLAT-1266). Deliberately a
+    # small set, distinct from the full ISO country_codes.COUNTRY_CODES list,
+    # so the per-country latency/TTFT/availability metrics stay cheap.
+    # Everything else clamps to "other" (see utils.clamp_launch_country).
+    # Grafana's country/region dropdown reads the same values.
+    MLPA_LAUNCH_COUNTRIES: set[str] = {"US", "CA", "FR", "DE"}
```
noahpodgurski: nit: Is it possible to reuse the data stored in `country_codes.py:COUNTRY_CODES`?

src/mlpa/core/metrics.py
```diff
@@ -72,14 +73,19 @@ def record_chat_availability_for(
     model: str,
     service_type: str,
     purpose: str,
+    client_country: str = "",  # "" (not yet authorized) clamps to "other" below
```
noahpodgurski: suggestion: We should just do `client_country: str` here to be consistent with the other params - it's already passed in with a default by the class request declaration

src/mlpa/core/prometheus_metrics.py
```diff
@@ -354,6 +357,34 @@ def build_metrics(registry: CollectorRegistry = REGISTRY) -> PrometheusMetrics:
...
+            "mlpa_chat_completion_latency_by_country_seconds",
+            "Chat completion latency in seconds, by launch-market client country "
+            "(AIPLAT-1266). Deliberately thin (no model/service_type/purpose) to "
+            "keep cardinality bounded; client_country is one of "
+            "LAUNCH_COUNTRIES or 'other'.",
+            ["result", "client_country"],
+            buckets=BUCKETS_COMPLETION,
+            registry=registry,
+        ),
+        chat_completion_ttft_by_country=Histogram(
```
noahpodgurski, line 370: question: Instead of adding these new metrics, would it make more sense to simply add `client_country` to the `chat_completion_latency` metric etc...?

src/mlpa/core/search.py
```diff
@@ -26,7 +26,10 @@ async def get_search(authorized_search_request: AuthorizedSearchRequest):
 async def _get_search(authorized_search_request: AuthorizedSearchRequest):
     start_time = time.perf_counter()
     body = sanitize_request_body(
-        authorized_search_request.model_dump(exclude_none=True)
+        authorized_search_request.model_dump(
+            exclude={"client_country"},
```
noahpodgurski: question: Do we also need to exclude `purpose` and `service_type` here?

src/tests/unit/test_authorize.py
```diff
@@ -65,6 +67,63 @@ async def test_authorize_search_request_returns_authorized_search_request(mocker
...
+        request=_make_request("/v1/chat/completions"),
+        chat_request=ChatRequest(
+            model="gpt-oss-120b", messages=[{"role": "user", "content": "hello"}]
+        ),
+        authorization="Bearer token",
+        service_type=authorize_module.ServiceType.ai,
+        purpose="chat",
+    )
+
+    assert result.client_country == ""
```
noahpodgurski, line 108: Does it only default to `other` if the client passes in an invalid (or non MLPA launch) country code?

subpath: Yes 👍

src/mlpa/core/prometheus_metrics.py
subpath: I think it will increase the dimensionality too much, currently  `chat_completion_latency` 1408 series, it will become 7k+ and will keep growing with each country...

src/mlpa/core/config.py
subpath: country_codes are too large of a list, I think we only care about several countries for now

src/mlpa/core/config.py
```diff
@@ -23,7 +24,18 @@ class Env(BaseSettings):
     # so the per-country latency/TTFT/availability metrics stay cheap.
     # Everything else clamps to "other" (see utils.clamp_launch_country).
     # Grafana's country/region dropdown reads the same values.
-    MLPA_LAUNCH_COUNTRIES: set[str] = {"US", "CA", "FR", "DE"}
+    # `NoDecode` + the validator below let ops set this as a plain
+    # comma-separated string (e.g. "US,CA,FR,DE,GB"); pydantic-settings'
+    # default complex-type decoding otherwise requires JSON syntax and would
+    # crash the app on startup for the intuitive comma-separated format.
+    MLPA_LAUNCH_COUNTRIES: Annotated[set[str], NoDecode] = {"US", "CA", "FR", "DE"}
```
noahpodgurski, line 31: question: Why the weird type here? Why not just declare it as a set?

src/mlpa/core/utils.py
```diff
@@ -95,6 +95,11 @@ def _clamp_to_set(
     return raw if raw in valid else fallback
 
 
+def get_client_country(request: Request) -> str:
+    """Read the raw, edge-stamped ``X-Geo-Country`` header. Clamp at metric time."""
+    return request.headers.get("X-Geo-Country") or ""
```
noahpodgurski, line 100: Does this work the same if the request passes in the header all lowercase? `x-geo-country`?

noahpodgurski (approved):
Left a few more Qs but looks good to me :)

src/mlpa/core/config.py
subpath: This is a pydantic shenanigans, because I've wanted that this field can be set from infra yamls

src/mlpa/core/utils.py
subpath: Yes 👍

PR #235, opened 2026-08-25
feat: privacy-filter support

noahpodgurski:
## What's new

- Add support for forwarding requests to https://github.com/Firefox-AI/privacy-filter via `POST /filter`
- Returns `privacy_filter: {version, model_id}` in `GET /health/readiness`
- Add config vars for privacy filter for url and master key
- Updates readiness LiteLLM version to name `litellm.version` (rather than `litellm.litellm_version`
- Update docs
- Add tests
- Sibling PR of https://github.com/mozilla/dataservices-infra/pull/2220
- [x] Validated in dev
https://mozilla-hub.atlassian.net/browse/AIPLAT-1260


src/mlpa/core/routers/filter/filter.py
```diff
@@ -0,0 +1,94 @@
...
+        client = get_http_client()
+        response = await client.post(
+            PRIVACY_FILTER_FILTER_URL,
+            headers=PRIVACY_FILTER_MASTER_AUTH_HEADERS,
+            json=authorized_filter_request.model_dump(
+                exclude={"user"}, exclude_none=True
+            ),
+            timeout=env.READINESS_CHECK_TIMEOUT_S,
+        )
+        return {
```
subpath, line 83: Suggestion: could we check the status code before returning? I think an upstream error would currently look like a 200.
```suggestion
        if response.status_code != 200:
            raise_and_log(response, service_type=...)
        return {
```

src/mlpa/core/routers/filter/filter.py
```diff
@@ -0,0 +1,94 @@
...
+        result = PrometheusResult.ERROR
+
+        client = get_http_client()
+        response = await client.post(
+            PRIVACY_FILTER_FILTER_URL,
+            headers=PRIVACY_FILTER_MASTER_AUTH_HEADERS,
+            json=authorized_filter_request.model_dump(
+                exclude={"user"}, exclude_none=True
+            ),
+            timeout=env.READINESS_CHECK_TIMEOUT_S,
```
subpath: Suggestion: I believe this timeout (2s) is meant for health pings. Could we give this call its own timeout?

src/mlpa/core/routers/health/health.py
```diff
@@ -52,13 +55,19 @@ async def _fetch_litellm_readiness(client):
...
+
+
 def _eval_litellm(litellm_http, version) -> tuple[bool, dict]:
     """Map the LiteLLM readiness result to (ready, sub-body).
 
     Only ready on a 200 with db connected and a healthy top-level status. A
     LiteLLM that is up but not ready can't serve MLPA traffic.
     """
-    unreachable = {"litellm_version": version, "status": "unreachable"}
+    unreachable = {"version": version, "status": "unreachable"}
```
subpath, line 70: Question: was renaming `litellm_version` to `version` intentional?

src/mlpa/core/routers/health/health.py
```diff
@@ -70,7 +79,31 @@ def _eval_litellm(litellm_http, version) -> tuple[bool, dict]:
...
+    return ready, {"version": version, **body}
+
+
+def _eval_privacy_filter(privacy_filter_http) -> tuple[bool, dict]:
+    """Map the Privacy Filter readiness result to (ready, sub-body).
+
+    Only ready on a 200 with db connected and a healthy top-level status. A
+    Privacy Filter that is up but not ready can't serve MLPA traffic.
+    """
+    unreachable = {"pf_version": "N/A", "status": "unreachable"}
```
subpath: Suggestion: I think `pf_version` vs `version` is used for the same field here, could we use one consistently?

src/mlpa/core/routers/filter/filter.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        return {
+            "results": response.json().get("results", []),
+            "model_id": response.json().get("model_id"),
+            "num_items": response.json().get("num_items", 0),
+        }
+    except Exception:
```
subpath: Suggestion: I think this except block just re-raises, could we add logging/metrics here like `raise_and_log`?

src/mlpa/core/auth/authorize.py
```diff
@@ -242,3 +244,19 @@ async def authorize_search_request(
...
+async def authorize_filter_request(
+    request: Request,
+    filter_request: PrivacyFilterRequest,
+    authorization: Annotated[str, Header()],
+) -> AuthorizedFilterRequest:
+    if not authorization:
+        raise HTTPException(status_code=401, detail="Missing authorization header")
+
+    fxa_user_id = await fxa_auth(authorization)
+    if not fxa_user_id or fxa_user_id.get("error"):
```
subpath, line 258: Suggestion: I believe this duplicates the FxA branch from `_authorize_common_request`, could we reuse it instead?

src/mlpa/core/routers/filter/filter.py
```diff
@@ -0,0 +1,94 @@
+import asyncio
```
subpath: Suggestion: I think a few unused imports here are left from the template, could we clean these up?

src/mlpa/core/routers/filter/filter.py
```diff
@@ -0,0 +1,94 @@
...
+        response = await client.post(
+            PRIVACY_FILTER_FILTER_URL,
+            headers=PRIVACY_FILTER_MASTER_AUTH_HEADERS,
+            json=authorized_filter_request.model_dump(
+                exclude={"user"}, exclude_none=True
+            ),
+            timeout=env.READINESS_CHECK_TIMEOUT_S,
+        )
+        return {
+            "results": response.json().get("results", []),
```
subpath: Suggestion: `response.json()` is called 3 times, maybe we could parse it once?
```suggestion
        data = response.json()
        return {
            "results": data.get("results", []),
            "model_id": data.get("model_id"),
            "num_items": data.get("num_items", 0),
        }
```

src/mlpa/run.py
```diff
@@ -136,6 +137,7 @@ async def get_metrics():
 app.include_router(appattest_router, prefix="/verify")
 app.include_router(play_router, prefix="/verify")
 app.include_router(user_router, prefix="/user")
+app.include_router(filter_router)
```
subpath, line 144: Suggestion: could we add `"Privacy Filter"` to `tags_metadata` for the docs?

src/mlpa/core/routers/health/health.py
noahpodgurski: Yeah I thought having the format of
```
{
"litellm": {
    "version": "x",
},
"privacy_filter": {
    "version": "x",
},
```
was better :)

src/mlpa/core/routers/filter/filter.py
noahpodgurski: good call 👍

src/mlpa/core/auth/authorize.py
noahpodgurski: That one depends on a service type whereas a filter request doesn't use a service type. Also common requests can include Play Integrity or App Attest auth but the filter request only uses fxa so I'd say they're different enough to stay separate

src/mlpa/run.py
noahpodgurski: Forgot about that, thanks!

src/mlpa/core/routers/filter/filter.py
noahpodgurski: We don't have a timeout defined for get completion so I'll just remove it

src/mlpa/core/routers/filter/filter.py
noahpodgurski: Will use `response.raise_for_status()` 👍

subpath (approved):

PR #249, opened 2026-09-01
feat: Added Liner service 

FieldMarshallObvious:
## What this PR does
Adds `liner-answer` service type, and restricts `liner-answers` model to only be allowed on that service type. New configurations where also added to support liner budgeting defaults, and liner-api-keys. Unit tests where added to cover the new set of cases for exa.

## QA Path
Verified on local sending requests with `liner-answers` model on liner-answer` service type and alternative service types.

**Valid Service Type:**
image

**Invalid Service Type:**
image


src/tests/unit/test_config.py
```diff
@@ -272,6 +312,9 @@ def test_valid_service_type_for_model_forced_pair():
     assert env.valid_service_type_for_model("answer", "exa") is True
     assert env.valid_service_type_for_model("ai", "exa") is False
     assert env.valid_service_type_for_model("search", "exa") is False
+    assert env.valid_service_type_for_model("liner-answer", "liner-answers") is True
```
noahpodgurski: I think having `liner-answer` and `liner-answers` is kind of confusing. Can we name the model just `liner`, and have the service-type just be `liner-answer`?

src/mlpa/core/config.py
```diff
@@ -98,6 +99,12 @@ def _parse_launch_countries(cls, raw: str | set[str]) -> set[str]:
     USER_FEATURE_BUDGET_ANSWER_TPM_LIMIT: int = 2000
     USER_FEATURE_BUDGET_ANSWER_BUDGET_DURATION: str = "1d"
 
+    USER_FEATURE_BUDGET_LINER_ANSWERS_BUDGET_ID: str = "linerAnswer"
```
noahpodgurski: Can we change this to be more in line with how the other budget ids are named?

litellm_config.yaml
```diff
@@ -19,6 +19,12 @@ model_list:
       model: openai/exa
       api_base: https://api.exa.ai
       api_key: os.environ/EXA_API_KEY
+  # Note this model requires a custom provider map to handle liner requests
+  #- model_name: liner-answers
+  #  litellm_params:
+  #    model: liner-answers/liner-quick-answer
```
noahpodgurski: the indents on these lines when uncommented are 1 space short

src/tests/unit/test_config.py
noahpodgurski: I could also see a case for it being under `answer`, like exa is

litellm_config.yaml
noahpodgurski: Also maybe add instructions (or commented out line) that to add the volume in `litellm_docker_compose.yaml` too

noahpodgurski (changes requested):
Super awesome that we have this working through LiteLLM now! Just a few comments 🙌

litellm_config.yaml
FieldMarshallObvious: Maybe this is my vim mode deceiving me but they seem to be correct?
image

src/mlpa/core/config.py
FieldMarshallObvious: Resolved in ee7df47dafb88f4e216ffc76e0ee8cc42a476203

src/tests/unit/test_config.py
FieldMarshallObvious: I changed the model name to `liner` then the service type to `liner-answer`. I think it probably makes the most sense to put it under answer but given that we would need to extend the `user_feature_budget` & associated path to parse by model.

FieldMarshallObvious: Resolved here d2c0cf9e211f2878031d67020297c1b4b1bfed43

litellm_config.yaml
noahpodgurski: Maybe it's just on my end then

src/mlpa/core/config.py
```diff
@@ -328,6 +343,7 @@ def forced_model_service_type_pairs(self) -> dict[str, list[str]]:
         return {
             "exa-search": ["search", "search-dev", "agent-search"],
             "exa": ["answer"],
+            "liner": ["liner-answer"],
```
noahpodgurski, line 347: I think we still want to keep `liner` and `exa` on the same `answer` service type. No need to create a new service type

FieldMarshallObvious: Talking with Steve about this, since there is a different budget for exa and liner this is a product call. Spoke with Sasha and we do not want to allow budget id to be selected by service_type, and model.

litellm_config.yaml
noahpodgurski: I select all lines, and do `CTRL+/` do you do that too or just delete the `#` on each line?

src/mlpa/core/config.py
FieldMarshallObvious: I spoke with Steve, and they would like to keep the service types separate to make the reporting more straight forward.

noahpodgurski: Sounds good, last I heard we wanted to combine the budgets but makes sense, let's wait to hear back from Steve

FieldMarshallObvious: Oh I spoke with Steve about this over slack when I sent this message.

noahpodgurski (approved):

subpath (approved):

PR #250, opened 2026-09-01
chore: update docker compose for privacy-filter

noahpodgurski:
## What's new

- Added privacy_filter to local testing setup
- Enabled by default and added disabling instructions with env `PRIVACY_FILTER_ENABLED=false`

### DIFF
--- a/src/mlpa/core/routers/filter/filter.py
+++ b/src/mlpa/core/routers/filter/filter.py
@@ -56,6 +56,12 @@ async def privacy_filter(
         AuthorizedFilterRequest, Depends(authorize_filter_request)
     ],
 ):
+    if not env.PRIVACY_FILTER_ENABLED:
+        raise HTTPException(
+            status_code=503,
+            detail={"error": "Privacy Filter is disabled."},
+        )
+
     user_id = authorized_filter_request.user
     if not user_id:
         raise HTTPException(
--- a/src/mlpa/core/routers/filter/filter.py
+++ b/src/mlpa/core/routers/filter/filter.py
@@ -56,12 +56,6 @@ async def privacy_filter(
         AuthorizedFilterRequest, Depends(authorize_filter_request)
     ],
 ):
-    if not env.PRIVACY_FILTER_ENABLED:
-        raise HTTPException(
-            status_code=503,
-            detail={"error": "Privacy Filter is disabled."},
-        )
-
     user_id = authorized_filter_request.user
     if not user_id:
         raise HTTPException(
--- a/src/mlpa/core/routers/health/health.py
+++ b/src/mlpa/core/routers/health/health.py
@@ -113,20 +113,17 @@ async def readiness_probe():
     # Run the checks concurrently. return_exceptions keeps one failure from
     # cancelling the rest, and folding the version fetch in here avoids a
     # separate serial round-trip.
-    (
-        litellm_ok,
-        app_attest_ok,
-        litellm_http,
-        litellm_version,
-        privacy_filter_http,
-    ) = await asyncio.gather(
+    checks = [
         litellm_pg.ping(),
         app_attest_pg.ping(),
         _fetch_litellm_readiness(client),
         get_litellm_version(client),
-        _fetch_privacy_filter_readiness(client),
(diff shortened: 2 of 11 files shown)


FieldMarshallObvious (approved):
Looks good to me!

start.sh
```diff
@@ -1,8 +1,31 @@
...
+if [ -f .env ]; then
+    set -a
+    # shellcheck disable=SC1091
+    . ./.env
+    set +a
+fi
+
+compose_profiles="${COMPOSE_PROFILES:-}"
+case "${PRIVACY_FILTER_ENABLED:-true}" in
+    true | TRUE | True | 1 | yes | YES | Yes)
```
subpath: Suggestion: I think this only matches a narrow set of truthy strings (`true`, `1`, `yes`), but Pydantic's bool parsing (used by `config.py`) accepts a wider set like `on`/`y`/`t`. Could we normalize the value first so both agree on what counts as true?
```suggestion
case "$(printf '%s' "${PRIVACY_FILTER_ENABLED:-true}" | tr '[:upper:]' '[:lower:]')" in
    true | 1 | yes | on | y | t)
```

src/mlpa/core/routers/filter/filter.py
```diff
@@ -56,6 +56,12 @@ async def privacy_filter(
         AuthorizedFilterRequest, Depends(authorize_filter_request)
     ],
 ):
+    if not env.PRIVACY_FILTER_ENABLED:
```
subpath: Suggestion: I think `Depends(authorize_filter_request)` still resolves (and hits FxA) before this check runs, since it's already a function parameter. Could we pull the flag check into its own dependency and list it before `authorize_filter_request`, so it short-circuits first?

start.sh
```diff
@@ -1,8 +1,31 @@
...
+        case ",${compose_profiles}," in
+            *,privacy-filter,*) ;;
+            *)
+                compose_profiles="${compose_profiles:+${compose_profiles},}privacy-filter"
+                ;;
+        esac
+        ;;
+esac
+
+COMPOSE_PROFILES=privacy-filter docker compose \
```
subpath: Suggestion: maybe use `${compose_profiles}` here too, so `down` and `up` stay in sync if another profile gets added later?
```suggestion
COMPOSE_PROFILES="${compose_profiles}" docker compose \
```

subpath (approved):

src/mlpa/core/routers/health/health.py
```diff
@@ -113,20 +113,17 @@ async def readiness_probe():
...
         get_litellm_version(client),
-        _fetch_privacy_filter_readiness(client),
-        return_exceptions=True,
-    )
+    ]
+    if env.PRIVACY_FILTER_ENABLED:
+        checks.append(_fetch_privacy_filter_readiness(client))
+
+    results = await asyncio.gather(*checks, return_exceptions=True)
+    litellm_ok, app_attest_ok, litellm_http, litellm_version = results[:4]
```
subpath: Suggestion: I think `results[:4]` / `results[4]` only stays correct because the privacy-filter check is always appended last. Could we unpack by name instead (e.g. zip against the check list, or a small dict), so a future added check can't silently shift the indices?

subpath:
Question: since Privacy Filter defaults to enabled but `privacy-filter:local` has to be built manually from a separate private repo, could this break `start.sh` out of the box for anyone who hasn't built that image yet? Maybe defaulting `PRIVACY_FILTER_ENABLED` to `false` locally (and documenting the opt-in) would be safer than opt-out.

noahpodgurski:
> Question: since Privacy Filter defaults to enabled but `privacy-filter:local` has to be built manually from a separate private repo, could this break `start.sh` out of the box for anyone who hasn't built that image yet? Maybe defaulting `PRIVACY_FILTER_ENABLED` to `false` locally (and documenting the opt-in) would be safer than opt-out.

Yep we can do that 👍

src/mlpa/core/routers/filter/filter.py
noahpodgurski: Great idea 🙌

PR #254, opened 2026-09-08
feat: add app_attest rollback script

subpath:
[AIPLAT-1189](https://mozilla-hub.atlassian.net/browse/AIPLAT-1189)

`migrate-app-attest-database.sh` only ever moves forward, `alembic upgrade head`, and aborts on any error. There was no way to undo a migration if a deploy needed rolling back. This adds `rollback-app-attest-database.sh` to actually run `alembic downgrade`, plus a rollback procedure doc and an expand/contract checklist so future migrations stay safe to leave applied if the code around them gets rolled back. Also fixes a timeout gap: the rollback session now overrides `statement_timeout`, since the role-level default (3s) is sized for cheap forward migrations, not a downgrade that might ALTER/DROP a column on a table that's grown large.

Part of a 3-PR set under AIPLAT-1189:
- mlpa-rollback job: https://github.com/mozilla/dataservices-infra/pull/2270


FieldMarshallObvious (approved):

docs/database-management.md
```diff
@@ -206,6 +206,40 @@ The `mlpa_user_capacity*` tables are created by migration, then reconciled on
...
+   rollback.
+2. Check whether the migration actually applied before the deploy failed:
+   `alembic -c alembic.ini -x sqlalchemy.url=... current`. If it's still on
+   the old revision, there's nothing to undo, stop here.
+3. If it did apply, this runs as a direct `kubectl apply` against the cluster,
+   bypassing ArgoCD (see `mlpa-rollback` in `dataservices-infra`), not
+   something your standing access should cover. Get temporary access first
+   with `mzcld jit elevate "AIPLAT-1189 rollback"` rather than reaching for a
+   standing admin grant.
+4. Run `scripts/rollback-app-attest-database.sh` against the same DB, with
```
FieldMarshallObvious: thought: Maybe we should also add steps for how to run the job on argo cd as well?
https://github.com/mozilla/dataservices-infra/pull/2270/changes#diff-9fda88243924e2477e9a73acd5491dca55a7048baf96f89f11d25ac74f480a13R416

FieldMarshallObvious: Since I believe the job would have to be done actually pointing mlpa to a previous version?

subpath: I've added the actual `helm template ... | kubectl apply` command inline in the doc

subpath: No, it doesn't need a specific previous-version image. Alembic migration files are append-only 👍 
I've added some info to the doc

docs/database-management.md
```diff
@@ -206,6 +206,52 @@ The `mlpa_user_capacity*` tables are created by migration, then reconciled on
...
+[AIPLAT-1189](https://mozilla-hub.atlassian.net/browse/AIPLAT-1189) covers the
+full context. `migrate-app-attest-database.sh` is forward-only, it runs
+`upgrade head` and aborts on any error, there's no downgrade path in that
+script. If a deploy needs to be rolled back and it included an app_attest
+migration:
+
+1. Revert the code on `main` (git revert + Argo sync), same as any other
+   rollback.
+2. Check whether the migration actually applied before the deploy failed:
+   `alembic -c alembic.ini -x sqlalchemy.url=... current`. If it's still on
```
noahpodgurski, line 220: question: Could we add this step to the script? It should be done manually ideally but it could be good as an extra check?

docs/database-management.md
```diff
@@ -206,6 +206,52 @@ The `mlpa_user_capacity*` tables are created by migration, then reconciled on
...
+   the old revision, there's nothing to undo, stop here.
+3. If it did apply, get temporary access first, this bypasses ArgoCD:
+   `mzcld jit elevate "AIPLAT-1189 rollback"` rather than reaching for a
+   standing admin grant.
```
noahpodgurski: question: Could you elaborate on this step?

docs/database-management.md
```diff
@@ -206,6 +206,52 @@ The `mlpa_user_capacity*` tables are created by migration, then reconciled on
...
+   to a specific revision if you need to go further:
+
+   ```
+   helm template k8s/llm-proxy -f k8s/llm-proxy/values-<env>.yaml \
+     --set mozcloud.tasks.jobs.mlpa-rollback.enabled=true \
+     --set-string mozcloud.tasks.jobs.mlpa-rollback.containers.mlpa-appattest-rollback.envVars.TARGET=-1 \
+     | kubectl apply -f -
+   ```
+
+   The job runs `scripts/rollback-app-attest-database.sh` under whatever
```
noahpodgurski, line 238: thought: Should be fine - we just have to be mindful with releasing changes to this script :)

docs/database-management.md
subpath: you mean what is mzcld or why additional permissions are needed?

https://github.com/mozilla/mozcloud/tree/main/tools/mzcld

I think we will not have permissions to kubectl apply directly against the prod cluster

noahpodgurski (approved):
LGTM 🚀

PR #259, opened 2026-09-10
feat: add redis service and RPM traffic contract config and tracking

noahpodgurski:
### tl;dr

- Added Redis service for fixed-window (every 60s) traffic contract enforcement
  - For managing traffic contracts with methods for checking and incrementing feature RPM and TPM.
  - Added separate Lua scripts for checking contract state and incrementing counters via Redis. I separated checking and incrementing since at the start of the request we don't know the token count (that info is in the response). So the order is 1. Check RPM/TPM limit, 2. Make request, 3. Increment RPM/TPM
  - Added testing for redis service and simple mocked contract testing
- **Over-contract requests are not rejected** The check returns `allowed=False` with `mode="borrowed"` or `mode="degraded"`, and logged, and are emitted on metric `requests_total` via `traffic_contract_rpm_mode` and `traffic_contract_tpm_mode`. No soft degradation is applied yet.
  - `normal`: feature and basket are within contract
  - `borrowed`: feature is over contract, but basket still has room
  - `degraded`: basket is over contract

### Redis
There's a new `redis_service` which uses the same Redis instance as LiteLLM, with the `mlpa:traffic_contract` key prefix. There's a discussion to be had around the current eviction policy and whether we should change it. Although we've had no evictions since inception, but it's something to keep in mind. It uses small Lua scripts to check fixed-window feature/basket state and increment RPM/TPM counters.

### Traffic contract design:

Traffic contracts are defined on a *per-feature* basis, so SW, S2S, etc... RPM and TPM limits are defined for each (ex: `SMART_WINDOW_TRAFFIC_CONTRACT_RPM_LIMIT`. RPM and TPM limits for all features (all requests/basket) combined is stored in `TOTAL_TRAFFIC_CONTRACT_RPM_LIMIT`. I also renamed `env.user_feature_budget` to `service_type_config` since it contains non budget info (feature info). Each budget/service type is combined into the `env.traffic_contract_config` which defines the RPM/TPM based off the service type's feature's definition. Not to be confused with the per user rpm/tpm limits defined in the `service_type_config (old user_feature_budget)`

The traffic_contract_enforcer runs as a handler in the request after the authorize_chat/search_request function has run. This is slightly less optimal (than in the middleware) but it keeps the code cleaner and reduces duplicated (service type, purpose, etc...) validation code and logic that fastapi already provides - plus it will make TPM tracking logic identical. A tradeoff with this is we do not count 401 requests towards the total.

After the upstream response, `redis_service.update_contracts` increments RPM and, when token usage is available, TPM.

<details>
<summary><h1>Detailed example flow (expand)</h1></summary>
The Redis key is a fixed-window bucket:

  `{TRAFFIC_CONTRACT_REDIS_KEY_PREFIX}:rpm:{bucket_start_epoch}`

  For example:

  `mlpa:traffic_contract:rpm:1789057680`

  The key is a Redis hash. Each feature gets its own field, and the shared basket
  uses `__basket__`.

  Example assumptions:

  - `TRAFFIC_CONTRACT_RPM_WINDOW_SECONDS=60`
  - `TRAFFIC_CONTRACT_TPM_WINDOW_SECONDS=60`
  - `TRAFFIC_CONTRACT_COUNTER_TTL_SECONDS=120`
  - `smart-window` feature RPM limit = `2`
  - `smart-window` feature TPM limit = `5000`
  - total basket RPM limit = `4`
  - total basket TPM limit = `10000`
  - `ai` and `memories` both map to `smart-window`
  - `s2s` maps to `s2s`
  - `s2s-android` maps to `s2s-android`
  - all requests below land in the same minute bucket
  - RPM and TPM are checked before the upstream request; RPM is incremented after the response, and TPM is incremented after the response when usage includes `total_tokens`

[Diagram link](https://mermaid.ai/live/edit#pako:eNqFkk1z4jAMhv8KozNp8UfIxyy9tLe9ZXraZidjsABPcUwdZynL8N_XxKELpNP6okjR6-eVkgMsjETIYbkxu8VaWDd6firrUXcK8lJCgW8j8mNu7x8atH_UAiO33-JMqBJ-nxt_nhob7dXRTtXS7GZBUVVz0byiq6q-oD1sVhurxaaTf5BoT6JDkkZtrMLmkkdvefSWR7_msZ7HvpuM3ZLYLYn9J82NtWaHsqtslFYOZTXfz5YoXGvx2gLvLfChhYZeTcs_9-C7hnvmX88d99D4U2gkammNkpfw-Dv4hW5op-Oc7wq2JK6skMMlBc21XTKKogf_e3WhoCELoWAhC6HgIQuhiEMWwxhWVknInW1xDBr9Rk4pHE6IEtwatf8suX-Uwr6WUNZHr9mK-pcx-iyzpl2tIV-KTeOzdiuFwycl_Bj6o2qxlmgfTVs7yJOUdZdAfoB3yGOW3WXJNJ4QPuUpZ-kY9pBPJ3dpwrOE0SwhJCP8OIa_HTW88CfJpmTCUkqO_wAwOyLb)

  This means the whole shared basket is now over contract. The request still
  continues for now, but this is the signal we can use later to degrade behavior:
  reduce max tokens, skip retries, route differently, or apply a short wait/jitter.

  When the next minute starts, Redis writes to a new key:

```
  mlpa:traffic_contract:rpm:1789057740
```

  The previous bucket stays around briefly because the TTL is longer than the
  window, which makes inspection/debugging easier without affecting enforcement.
</details>

### If not enforcing limits, why not just create alerts based on this RPM/TPM data we already have in Grafana?

Even though we're not enforcing a 429 on requests that go over the traffic contract, the request metric now includes `traffic_contract_rpm_mode` and `traffic_contract_tpm_mode`, which provides insight into the percentage of traffic is `normal`, `borrowed`, or `degraded` and tune contract numbers from there.

https://mozilla-hub.atlassian.net/browse/AIPLAT-1060
https://mozilla-hub.atlassian.net/browse/AIPLAT-1064


src/mlpa/core/config.py
```diff
@@ -107,104 +113,134 @@ def _parse_launch_countries(cls, raw: str | set[str]) -> set[str]:
...
             "answer": {
+                "feature": self.FEATURE_SMART_WINDOW,
                 "budget_id": self.USER_FEATURE_BUDGET_ANSWER_BUDGET_ID,
+                "budget_duration": self.USER_FEATURE_BUDGET_ANSWER_BUDGET_DURATION,
                 "max_budget": self.USER_FEATURE_BUDGET_ANSWER_MAX_BUDGET,
                 "rpm_limit": self.USER_FEATURE_BUDGET_ANSWER_RPM_LIMIT,
                 "tpm_limit": self.USER_FEATURE_BUDGET_ANSWER_TPM_LIMIT,
-                "budget_duration": self.USER_FEATURE_BUDGET_ANSWER_BUDGET_DURATION,
             },
             "sw-answer": {
```
subpath, line 250: Suggestion: I think `sw-answer` is missing the `feature` key , I think it could be a   `KeyError('feature')` 🤔

src/mlpa/core/services/redis_service.py
```diff
@@ -0,0 +1,320 @@
...
+            ttl_seconds,
+            increment_amount,
+        )
+
+        return TrafficContractCounters(
+            feature_count=int(result[0]),
+            basket_count=int(result[1]),
+        )
+
+    async def check_feature_rpm(
```
subpath: Question: I think `check_feature_rpm`, `increment_feature_rpm`, `check_feature_tpm`, and `increment_feature_tpm`  are not used inside unit tests, is it correct?

src/mlpa/core/completions.py
```diff
@@ -330,6 +333,9 @@ async def _read_next_chunk(
             authorized_chat_request, result, time.perf_counter() - start_time
         )
         record_chat_availability(authorized_chat_request, availability_reason)
+        await redis_service.update_contracts(
```
subpath: Suggestion: 

maybe wrap it in try except
```
finally:
        record_completion_latency(
            authorized_chat_request, result, time.perf_counter() - start_time
        )
        record_chat_availability(authorized_chat_request, availability_reason)
        try:
            await redis_service.update_contracts(
                service_type=authorized_chat_request.service_type, usage=usage
            )
        except Exception as exc:
            logger.error(f"Traffic contract update failed: {exc}")
            if not env.TRAFFIC_CONTRACT_FAIL_OPEN_ON_REDIS_ERROR:
                raise
```

src/mlpa/run.py
```diff
@@ -82,6 +68,10 @@ async def lifespan(app: FastAPI):
         await app_attest_pg.connect()
         app_attest_connected = True
 
+        if env.ENABLE_TRAFFIC_CONTRACT_ENFORCEMENT:
+            await redis_service.connect()
```
subpath, line 72: Comment: so now redis service will become a hard dependency, without it MLPA will not start

src/mlpa/core/services/redis_service.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @classmethod
+    def retry_after_seconds(
+        cls, *, now: int | None = None, window_seconds: int = 60
+    ) -> int:
+        current_time = int(time.time()) if now is None else now
+        elapsed_in_bucket = current_time % window_seconds
+        return window_seconds - elapsed_in_bucket
+
+    async def check_feature_traffic_contract(
```
subpath: Suspicious: 
I think we can have a race condition ...

so I understood it as: 
1. requests comes and we run HGET to read current count, predicts count + 1
2. we check if to allow the request or not
3. we don't persists this  predicts count + 1 value
4. After LLM responds, we run increment_feature_traffic_contract runs and it does HINCRBY - pushing the value to Redis

but during the duration of the LLM response more calls will come...

Maybe we should persist the value on the before LLM call? 🤔 
But for TPM calls it will not work

src/mlpa/run.py
```diff
@@ -218,6 +161,7 @@ async def chat_completion(
         service_type=authorized_chat_request.service_type,
         model=authorized_chat_request.model,
     )
+    await enforce_traffic_contract(request, authorized_chat_request.service_type)
```
subpath, line 175: Suggestion:  maybe register `enforce_traffic_contract` into `MIDDLEWARE_EXECUTION_ORDER`

src/mlpa/core/middleware/instrumentation.py
```diff
@@ -13,6 +14,22 @@
 )
 
 
+def _traffic_contract_mode_label(
```
subpath, line 17: Suggestion: I think `_traffic_contract_mode_label` duplicates what `_clamp_to_set` already does  🤔 
```suggestion
def _traffic_contract_mode_label(
    request: Request,
    key_type: TrafficContractKeyType,
) -> str:
    mode = getattr(
        request.state,
        f"traffic_contract_{key_type.value}_mode",
        "N/A",
    )
    if isinstance(mode, TrafficContractMode):
        mode = mode.value
    return _clamp_to_set(mode, {m.value for m in TrafficContractMode}, "N/A")
```

src/mlpa/core/middleware/traffic_contract_enforcer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if not env.ENABLE_TRAFFIC_CONTRACT_ENFORCEMENT:
+        return
+
+    contract = env.traffic_contract_config.get(service_type)
+    if contract is None:
+        return
+
+    try:
+        rpm_decision, tpm_decision = await asyncio.gather(
```
subpath: suggestion: maybe get these values in 1 trip? 

lua script could be something like (I didn't test it)

```lua
local function check(key, feature_field, basket_field, feature_limit, basket_limit, inc_amount)
    if basket_limit == 0 and feature_limit == 0 then
        return {1, 0, 0, "", "normal", ""}
    end
    local feature_count = tonumber(redis.call("HGET", key, feature_field) or "0") + inc_amount
    local basket_count = tonumber(redis.call("HGET", key, basket_field) or "0") + inc_amount
    if basket_limit > 0 and basket_count > basket_limit then
        return {0, feature_count, basket_count, "basket", "degraded", tostring(basket_count/basket_limit)}
    end
    if feature_limit > 0 and feature_count > feature_limit then
        return {0, feature_count, basket_count, "feature", "borrowed", tostring(feature_count/feature_limit)}
    end
    return {1, feature_count, basket_count, "", "normal", ""}
end

local rpm = check(KEYS[1], ARGV[1], ARGV[2], tonumber(ARGV[3]), tonumber(ARGV[4]), tonumber(ARGV[5]))
local tpm = check(KEYS[2], ARGV[6], ARGV[7], tonumber(ARGV[8]), tonumber(ARGV[9]), tonumber(ARGV[10]))
return {rpm, tpm}

```

src/mlpa/core/services/redis_service.py
noahpodgurski: I had tests for them but you're right these funcs are outdated with `increment_feature_traffic_contract` now

src/mlpa/core/completions.py
noahpodgurski: forgot to add handling here, thank you 🙏

noahpodgurski: I added the handling inside `update_contracts` and also converted the function to be ran as a task
```
        asyncio.create_task(
            redis_service.update_contracts(
                service_type=authorized_chat_request.service_type,
                usage=usage,
            )
        )
```
otherwise it blocks the data from being returned, so it's more efficient 👍

src/mlpa/run.py
noahpodgurski: Only if `ENABLE_TRAFFIC_CONTRACT_ENFORCEMENT=true` yes?

But LiteLLM won't work without it either

src/mlpa/run.py
noahpodgurski: I added a bit about this in the description 🙏 

> The traffic_contract_enforcer runs as a handler in the request after the authorize_chat/search_request function has run. This is slightly less optimal (than in the middleware) but it keeps the code cleaner and reduces duplicated (service type, purpose, etc...) validation code and logic that fastapi already provides - plus it will make TPM tracking logic identical. A tradeoff with this is we do not count 401 requests towards the total.

src/mlpa/core/middleware/traffic_contract_enforcer.py
noahpodgurski: Yeah I was trying to keep the code modular but combining it into one request it just more optimal

src/mlpa/core/services/redis_service.py
noahpodgurski: That was my first approach, to check and increment in the same script. But like you said for TPM it doesn't work, that's why I split them both out. I figured it's better to have them unified in that way than for just the RPM check and increment be persisted pre-request, even if we have the capability for it

subpath (approved):

PR #261, opened 2026-09-11
ci: add Claude-written narrative to release notes

jgauf:
## What

The release notes page listed which tickets and PRs shipped, but said nothing about what the release actually *did*. This adds a Claude-written narrative above those lists: a release summary, then per-theme sections covering **what was delivered**, **why it mattered**, and **which epic/theme it supports**.

The flat Jira-ticket and pull-request lists stay underneath, unchanged, for traceability.

The one-line summary is also exposed as a step output, so both Slack notifications lead with what shipped instead of a bare link:

```
🚀 MLPA v1.0.36 released
This release focuses on per-country request metrics and completion
budget enforcement, with the remainder being dependency upkeep.
<https://mozilla-hub.atlassian.net/wiki/…|View release notes>
```

## How

- **`.github/scripts/release_narrative.py`** (new) — flattens Jira ADF descriptions to text, builds the context block (per ticket: key, summary, type, epic, labels, description; per PR: title, body, ticket keys), and calls `claude-opus-5` with adaptive thinking and a `json_schema` structured output, so the response is validated rather than prose to be parsed. Per-item caps of 2000 chars keep a long writeup from blowing up the request.
- **`release_to_jira.py`** — new step 5b fetches each attached ticket's fields. Epics come from `parent`; if AIPLAT is on the old hierarchy it discovers the classic *Epic Link* custom-field id from `/rest/api/3/field` rather than hardcoding a per-tenant id. Epic summaries are fetched once and cached.
- **`DRY_RUN=1`** reads from GitHub and Jira and generates the narrative, but writes nothing — prints the page HTML and exits. For iterating on the prompt locally.

Two things worth a reviewer's attention:

- **Fail-open.** A missing `ANTHROPIC_API_KEY`, a rate limit, a refusal, or a malformed response logs to stderr and publishes the page with just the lists. The API can never block a release.
- **Anti-inflation prompting.** The system prompt requires every claim to be grounded in the ticket/PR data and prefers a plain description over invented impact. With inputs as thin as `chore: bump version to 1.0.32`, an unconstrained model confabulates business value — which is exactly what makes generated release notes untrustworthy.

Also replaces the webhook step's `sed`-based JSON escaping with `json.dumps`, and passes interpolated values via `env:` rather than inlining `${{ }}` in `run:`. The old escaping covered backslashes and quotes only, so a backtick or `$(...)` would have been shell-evaluated — worth closing now that the text is model-generated.

## Before merging

- [ ] `gh secret set ANTHROPIC_API_KEY -R Firefox-AI/MLPA` (Claude service-account key)
- [x] Local `DRY_RUN=1` run against a recent tag to review real generated output
- [ ] Decide whether `effort: high` is warranted, or step down to `medium`

## Testing

Verified live: a `DRY_RUN=1` run against v1.0.35 generated the narrative end-to-end against the real API, writing nothing to Jira or Confluence. Also covered: ADF flattening on nested documents, context building, HTML escaping (`<`, `>`, `&`, `"`), the no-epic theme, the empty-narrative and no-key fail-open paths, `$GITHUB_OUTPUT` sanitization against multiline/oversized/quote-bearing summaries, and the webhook payload against a summary containing quotes, backticks, `$(whoami)` and backslashes. The `DRY_RUN` path was run end-to-end against a stubbed HTTP layer rigged to raise on any non-GET; it exits 0 with no writes attempted.

### Note for local testing

If you hit `APIConnectionError: Connection error.` locally, check your Brotli version before suspecting the network or your key. Anaconda ships Brotli 1.0.9, whose `Decompressor.process()` predates the `output_buffer_limit` keyword that httpx2 2.12 passes when decoding the API's `content-encoding: br` responses — the `TypeError` surfaces as a bogus connection error. A clean venv (`pip install requests anthropic`) avoids it, and matches what CI builds.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


noahpodgurski (approved):
LGTM, let's give it a shot

PR #262, opened 2026-09-11
feat: distinguish MLPA user budget 429 from LiteLLM global budget 429

noahpodgurski:
## What's new

Adds 
```
500 {
  "detail": {
    "error": 10
  }
}
```

when LiteLLM global budget is hit, rather than returning typical 429 user out of budget response.

Also counts as `Availability Failure` for metrics

https://mozilla-hub.atlassian.net/browse/AIPLAT-1212


src/mlpa/core/errors.py
```diff
@@ -46,29 +48,43 @@ def availability_reason(self) -> AvailabilityReason:
...
     ),
     ERROR_CODE_UPSTREAM_RATE_LIMIT_EXCEEDED: (
+        429,
         PrometheusRejectionReason.RATE_LIMITED,
         "60",
         "Upstream rate limit exceeded",
     ),
 }
 
+_LITELLM_GLOBAL_BUDGET_ERROR = "ExceededBudget: User=default_user_id over budget"
```
subpath, line 78: Question: this only matches the `User={id}` format. LiteLLM also raises budget errors as `End User=`, `Key over`, and `Team=`, which would fall through to the regular per-user 429 path instead. Could we match all four formats?

src/mlpa/core/config.py
```diff
@@ -576,6 +578,11 @@ def __init__(self):
                         "value": {"error": ERROR_CODE_BUDGET_LIMIT_EXCEEDED},
                         "description": "Budget limit exceeded. Check Retry-After header (86400 seconds = 1 day).",
                     },
+                    "global_budget_exceeded": {
+                        "summary": "Global budget limit exceeded",
+                        "value": {"error": ERROR_CODE_GLOBAL_BUDGET_LIMIT_EXCEEDED},
```
subpath: Suggestion: I believe this OpenAPI example still describes the old 429/86400s behavior, but the code now returns 500 with a 300s Retry-After for this error code. Could we update the description to match?

src/mlpa/core/errors.py
```diff
@@ -46,29 +48,43 @@ def availability_reason(self) -> AvailabilityReason:
...
+
 
 def _parse_rate_limit_error(error_text: str) -> int | None:
     if not error_text:
         return None
     try:
+        if _LITELLM_GLOBAL_BUDGET_ERROR in error_text:
```
subpath: Suggestion: maybe reuse `is_rate_limit_error`'s matching here instead of a separate substring check?

src/tests/unit/test_completions.py
```diff
@@ -651,6 +652,48 @@ async def test_get_completion_budget_limit_exceeded_400(mocker, metrics_spy):
     assert _rejection_count(metrics_spy, PrometheusRejectionReason.BUDGET_EXCEEDED) == 1
 
 
+async def test_get_completion_global_budget_limit_exceeded(mocker, metrics_spy):
```
subpath, line 659: Suggestion:
maybe we could use a shared fixture here? Because i think this mock `ExceededBudget` text is duplicated across `test_completions.py`, `test_precompletion_availability.py`, and `test_search.py`

src/mlpa/core/errors.py
noahpodgurski: `ExceededBudget: User=default_user_id over budget` is the trigger for global budget 429.

Our other 429 error parsing already handles the per user 429 path 👍

src/mlpa/core/config.py
noahpodgurski: Yes good call

src/mlpa/core/errors.py
noahpodgurski: I thought it'd be more efficient to do a string check for this one and returning early - instead of doing the json.loads but I guess the typical path would run through it anyway so it makes sense to move it

subpath (approved):

PR #263, opened 2026-09-14
test(litellm-contract): AIPLAT-910 upgrade safety

subpath:
[AIPLAT-910](https://mozilla-hub.atlassian.net/browse/AIPLAT-910)

First step of the LiteLLM upgrade contract test. Two things:

1. Dependabot now watches the pinned LiteLLM image tag in `litellm_docker_compose.yaml` and opens a PR when a new version is out.
2. That PR runs a new test that checks LiteLLM's own Prisma migrations applied cleanly on boot

More checks (auth accept/reject, full request-shape coverage, budget-reset behavior, MLPA's direct-SQL writes into LiteLLM's tables) are coming in follow-up PRs.


src/tests/integration/test_litellm_contract.py
```diff
@@ -0,0 +1,110 @@
...
+            try:
+                return await conn.fetch(
+                    "SELECT migration_name, finished_at, rolled_back_at "
+                    'FROM "_prisma_migrations"'
+                )
+            finally:
+                await conn.close()
+
+        rows = asyncio.run(_fetch())
+        assert rows, "no rows in _prisma_migrations -- LiteLLM's schema never migrated"
```
noahpodgurski, line 103: nit: Could we add detection for *any* failed migration? Or if we could validate the latest migration aligns with what's expected from LiteLLM's source that'd be a great addition too

noahpodgurski (approved):

src/tests/integration/test_litellm_contract.py
subpath: yes, it's a quick start example, I've made a lot of tests, then realized that diff is too large, so I split it into PRs, I will submit next PR soon 👍

PR #274, opened 2026-09-17
feat(deps): upgrade pyattest and cbor2

subpath:
## Summary
- Bumps `pyattest` 1.0.2 -> 1.0.5, `cbor2` 5.7.0 -> 6.1.3.
- Unblocks the cbor2 6.x bump (#207): pyattest 1.0.2 imports cbor2's C extension from a path removed in 6.x. 1.0.5 fixes it.
- pyattest 1.0.5 made `Attestation.verify()` a coroutine, updated `verify_attest()` for that.

## Threading
- `run_in_threadpool(attestation.verify)` (old code, unchanged) -> runs but verifies nothing: constructs the coroutine in the thread, never drives it.
- Can't drop `run_in_threadpool` either: `verify()` is CPU-bound, so awaiting it directly blocks the event loop for every concurrent request.
- I think the best option is: `run_in_threadpool(asyncio.run, attestation.verify())`.

## Benchmark: real `verify_attest()`, main vs this branch
400 concurrent calls, 3 runs each, DB write and GCS cert download mocked:

| | Throughput (req/s) |
|---|---|
| main | 603 / 589 / 634 (avg 609) |
| this branch | 670 / 697 / 704 (avg 690) |

~13% faster, no regressions.


src/mlpa/core/routers/appattest/appattest.py
```diff
@@ -28,7 +29,9 @@
 PROJECT_ROOT = Path(__file__).resolve().parents[5]
 
 
-@lru_cache(maxsize=1)
+@lru_cache(
+    maxsize=2
```
noahpodgurski: thought: `APP_ATTEST_QA` is false in prod, so this optimization wouldn't affect prod

noahpodgurski (approved):
Nice! 🚀

